PR #4, opened 2021-12-16
Webpack 5, finish es6 module convert

Cellivar:
Does what it says on the tin.

To properly handle ES6 modules we needed to move to webpack 5, which was naturally an entire can of worms on its own.

This updates a fair number of dependencies to get them to play nice together, and tweaks some of the import syntax to properly run as ES6.

On my machine `npm run build`, `npm run watch`, and `npm run test` are all three succeeding and the extension appears to function normally in my browser. I think it's actually fixed this time.

To test this I've messed around with importing my LP2844 libraries and I stopped getting the previous errors about ES6 things, so I think this resolved the rest of those problems and we can get down to brass tacks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,23 +21,23 @@
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
     "chrome-extension-cli": "^0.2.3",
-    "copy-webpack-plugin": "^6.4.1",
-    "css-loader": "^4.3.0",
+    "copy-webpack-plugin": "^10.1.0",
+    "css-loader": "^6.5.1",
     "file-loader": "^6.2.0",
-    "mini-css-extract-plugin": "^0.10.1",
+    "mini-css-extract-plugin": "^2.4.5",
     "mocha": "^9.1.3",
     "mock-local-storage": "^1.1.19",
     "msw": "^0.36.3",
-    "node-fetch": "^2.6.6",
+    "node-fetch": "^3.1.0",
     "sinon": "^12.0.1",
     "size-plugin": "^2.0.2",
-    "webpack": "^4.46.0",
-    "webpack-cli": "^3.3.12",
+    "webpack": "^5.65.0",
+    "webpack-cli": "^4.9.1",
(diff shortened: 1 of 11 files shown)


src/regfox/regfox_api.js
```diff
@@ -1,12 +1,11 @@
-const { data } = require('jquery');
```
kobaj, line 1: I have no idea how that got in there, thank you for cleaning it up!

test/regfox/regfox_api.test.js
```diff
@@ -1,17 +1,19 @@
...
+import chaiAsPromised from 'chai-as-promised';
 
-chai.use(chaiAsPromised);
-chai.should();
+use(chaiAsPromised);
+should();
 const { expect } = chai;
 
-const sinon = require("sinon");
+import { stub } from "sinon";
```
kobaj, line 8: That's fair, yeah, I should have been better about not importing the entirety of sinon and chai. :)

package.json
```diff
@@ -21,23 +21,23 @@
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
     "chrome-extension-cli": "^0.2.3",
-    "copy-webpack-plugin": "^6.4.1",
-    "css-loader": "^4.3.0",
+    "copy-webpack-plugin": "^10.1.0",
```
kobaj, line 24: Oh awesome! Glad you went through and grabbed the latest dependencies!

kobaj (approved):

test/regfox/regfox_api.test.js
Cellivar: A _solid_ 50% of this commit is me just doing what Visual Studio Code told me to do and then trying to figure out why it did that. This was one of them, where it helpfully discovered that only the `stub` method was used and swapped out the imports. Kinda neat!

package.json
Cellivar: Mostly a requirement. I played this game called "update one thing and figure out what else broke". I stopped when stuff went green.

PR #5, opened 2021-12-16
Add printing from RegDesk page

Cellivar:
Does what it says on the tin.

* Adds bootstrap and a dark theme to keep my eyes from falling out.
* Adds the js infrastructure around the RegDesk page.
* Updates the WebZLP library to play nicer after I fixed a few mistakes with the ES6 syntax in my library.
* Adds the appropriate infrastructure to connect to and run test prints from printers.


Cellivar:
Basic testing UI lifted out of my other project and dropped in here. This won't last long but is useful for testing!

image

public/regdesk.html
```diff
@@ -4,10 +4,28 @@
...
+  <meta name="color-scheme" content="light dark">
+  <meta name="theme-color" content="#111111" media="(prefers-color-scheme: light)">
+  <meta name="theme-color" content="#eeeeee" media="(prefers-color-scheme: dark)">
+  <link rel="stylesheet" href="regdesk.css" />
 </head>
 
 <body>
-    <h1>This will be an awesome page with stuff!</h1>
+  <div class="container-fluid">
+      <div class="row" style="padding-top: 16px;">
```
kobaj: Inline styles are considered a poor practice[1], and specifying pixels is also fallen out of fashion[2]. 

[1] https://stackoverflow.com/questions/2612483/whats-so-bad-about-in-line-css
[2] https://stackoverflow.com/a/43131958/1204809


This is all temporary code, so I'm not gunna block on it by any means. But figured you might be interested in learning that.

public/regdesk.html
```diff
@@ -4,10 +4,28 @@
...
+          <div class="col-3">
+              <div class="d-grid gap-2 col-8 mx-auto mb-3">
+                  <button class="btn btn-primary" id="addprinter">➕ Add Printer</button>
```
kobaj: Oh man I'm torn. Usually you want to avoid emoji characters because they will render differently on different devices (and sony is particularly ugly). Usually you would use something like icons instead [1].

[1] https://fonts.google.com/icons?selected=Material+Icons+Outlined 

But we are guaranteed to always be on chrome, so its not that big of a deal. (And again, temporary code, so not gunna block by any means).

public/regdesk.html
```diff
@@ -4,10 +4,28 @@
...
+              </div>
+              <ol id="printerlist" class="list-group list-group-numbered">
+              </ol>
+          </div>
+          <div class="col-9">
```
kobaj: Extra empty div?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+
+    #printerType = LP2844;
+
+    constructor() {
+    }
+
+    get printers() { return this.#printers; }
+
+    async printerConnect(e) {
+        if (!this.#devices.includes(e.device)) {
```
kobaj: I'm a fan of de-nesting if statements. So that this would become

```
if(this.#devices.includes(e.device) {
  return;
}
this.#devices.push(e.device);
...
```

https://softwareengineering.stackexchange.com/questions/47789/how-would-you-refactor-nested-if-statements

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+    secondaryPrinter = 1;
+
+    #printerType = LP2844;
+
+    constructor() {
+    }
+
+    get printers() { return this.#printers; }
+
+    async printerConnect(e) {
```
kobaj: Usually methods are verbs, so "connectPrinter", unless these are meant to be event driven in which case maybe "onPrinterConnect"? I'm not super picky admittedly.

Edit: Ahh, I think I see now. This is both in response to a printer connecting, and also does the connection. That feels kinda funny, but I can't think of a better way to do it. Maybe there needs to be a distinction between "onUsbDeviceConnected" (like plugged in) and "nowConnectThePrinter" (actually initialize the thing).

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+        if (!this.#devices.includes(e.device)) {
+            this.#devices.push(e.device);
+            let printer = new this.#printerType(e.device);
+            await printer.connect();
+            this.#printers.push(printer);
+            this.primaryPrinter = 0;
```
kobaj: This line doesn't do anything.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
+import { LabelEpl, LP2844 } from 'WebZLP/src/LP2844'
+
+export class PrinterManager {
+    #devices = [];
+    #printers = [];
```
kobaj: Could make this a map? Or at least a list of objects? Would ensure you never have the two get out of order relative to each other.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    async printerDisconnect(e) {
+        const idx = this.#devices.findIndex(i => e.device == i);
+        if (idx > -1) {
+            const p = this.#printers[idx];
+            this.#devices.splice(idx, 1);
+            this.#printers.splice(idx, 1);
+            this.resetPrinterButtons();
```
kobaj: I had a huge comment typed up, but after thinking about it this may actually just be a typo.

(If it isn't a typo, oh boy, we got some fun ahead of us :D).

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+            let device = await navigator.usb.requestDevice({
+                filters: [
+                    { vendorId: this.#printerType.usbVendorId }
+                ]
+            });
+
+            await this.printerConnect( { device : device} )
+        } catch (e) {
+            console.log("Failed to connect to printer!");
+            console.log(e);
```
kobaj: should this be console.error?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+            return;
```
kobaj: Should this return false or something? I assume that the client needs to know that pair failed?

src/regdesk/regdesk.js
```diff
@@ -0,0 +1,106 @@
...
+var printerMgr;
+
+document.addEventListener('readystatechange', async () => {
+    if (document.readyState === "complete") {
+
+        const printerList = document.getElementById("printerlist");
+        printerMgr = new PrinterManager();
+
+        document.getElementById("addprinter").addEventListener('click', async () => {
+            await printerMgr.printerPair();
```
kobaj: This feels a little disjointed. Like there are two mechanisms to establishing communication with a printer. The navigator.usb.addEventListener (automatic? physical?) and this printerPair (manual? configuration?) and they behave differently.

Put another way, its almost like the "old" way of inserting a floppy. You would do the physical action of insertion, then go to your command prompt and type "mount  /mnt/floppy". Compared to today, where you do the physical action of insertion, then your pc pops up with an autoplay dialog. The latter is a lot nicer.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    async printerReload(e) {
```
kobaj: I don't think you do anything with e? And it looks like this reloads all printers (plural)?

EDIT: Looks like this method isn't called, which is good because it didn't sit well with me haha, can it be deleted?

public/regdesk.html
```diff
@@ -4,10 +4,28 @@
 <head>
   <meta charset="UTF-8">
   <title>FC RegDesk</title>
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <meta name="color-scheme" content="light dark">
```
kobaj, line 8: Heck yes! I love dark theme, thank you so much for including it up front :D.

src/regdesk/regdesk.js
```diff
@@ -0,0 +1,106 @@
...
+            resetPrinterButtons(printerMgr);
+        });
+
+        navigator.usb.addEventListener('connect', async (e) => printerMgr.printerConnect(e));
```
kobaj: This is so cool learning about usb interaction in the browser! Thank you for taking the initiative to learn the hard parts and make it look easy for me :D.

Cellivar:
All above are good points, in light of "let's be less shenanigan-y" I'm going after cleaning up some of this code now instead of later.

The printer interface and code should be more targeted towards the purpose: 1 or 2 printers for a reg interface. To that end I've slapped this together and I started filling it out before I fell asleep at my desk for a bit. Here's what it looks like (code isn't ready).

Let me know what you think, tomorrow I'll have more time to flesh out at least the part where the printers can be set up and that'll wrap up this PR so we can move onto the interface for dumping data into the printers for printing.

image

kobaj:
Thats a really good looking screenshot, I love the colors.

I do have a few concerns:
1. Are the accept cash/cards buttons new features? The original design document didn't mention accepting payment.
2. It looks like a lot of clicks and configuration is available? Again the original design document specified speed and simplicity.

Cellivar:
> Are the accept cash/cards buttons new features? The original design document didn't mention accepting payment.

Yes, thank you for the reminder that I need to update that doc. While talking with Woofy during inventory she mentioned that a significant fraction of folks doing on-site are paying in cash instead of ahead of time with a card. I threw them in there as an experiment with a toggleable button that can enable or disable the station accepting payments. This would then influence steps in the workflow.

For example, if a desk station wasn't accepting payments and someone's registration came up as unpaid, there would be a popup saying "Payment incomplete. Direct person to payment line" for the cashier. If the desk station had payments enabled it would instead instruct the cashier to accept payment and offer a big 'ol button to mark it as paid.

I was planning to tackle the UI nuances there next week, so I'll back those buttons out again for now.

> It looks like a lot of clicks and configuration is available? Again the original design document specified speed and simplicity.

Less than it looks at first glance. The printers have a "connect" button and then some debugging buttons like printing a test page or feeding a label. The accept payment buttons are toggles, and the "Connected" status indicator in the top left is just a passive display of whether RegFox's API is responding and being kept alive by your shenanigans.

If we get to the point where I can't explain the entire cashier-relevant interface in a 24pt flyer with diagrams that's kind of my mental tipping point on "this is too complex".

public/regdesk.html
Cellivar: I was vaguely aware that it was frowned upon but not the deeper reasons, that was useful, thanks! Resolving as I found the more correct answer for properly spacing elements and there are no inline styles for spacing anymore.

public/regdesk.html
Cellivar: In the tradeoff of emoji variance vs adding more dependencies to the project I think we'll be okay with emoji. We aren't really using it as part of a coherent theme so much as using them for their ideographic meaning. That should remain consistent through time (even if the style changes drastically) so I'm generally okay with emoji in my functional systems.

And as you pointed out we're targeting a single browser, so it's less a concern here. WebUSB is only functional in Chromium based browsers anyhow.

public/regdesk.html
Cellivar: Placeholder, now moot.

src/regdesk/printer_manager.js
Cellivar: Copy/paste error specifically, yep.

src/regdesk/printer_manager.js
Cellivar: Yes, and everything should probably return something more useful than nothing. I'll do a cleanup pass to make sure the interface surface is a little more coherent.

src/regdesk/printer_manager.js
Cellivar: Internet says [handleConnect](https://stackoverflow.com/a/60052668/1309423) would be best, I'll roll with that.

The lifecycle is dependent on whether the user has allowed this printer in the past. 

The normal lifecycle is that we ask Chrome to go looking for a USB device filtered by a manufacturer's ID (Zebra, in our case) and the user gets a little popup asking them to select what device they'd like the page to talk to. Once the user selects a device that's what we get returned from the `requestDevice` method.

If they have previously paired the printer (and the printer doesn't have any weirdness going on like a missing serial number) then Chrome can skip the pairing permission process and just fire off the `connect` event to tell us the device reappeared for us to play with.

Since we've finished the `requestDevice` pairing workflow and are now holding a device just like if the `connect` event fired, might as well just turn around and hand it off. It's the same codepath.

src/regdesk/printer_manager.js
Cellivar: Cleaned up 👍🏻

src/regdesk/printer_manager.js
Cellivar: Took a slightly different approach, let me know what you think.

src/regdesk/printer_manager.js
Cellivar: Yep, normalized everywhere.

src/regdesk/printer_manager.js
Cellivar: Cleaned up a bit. The original method here grew very organically over time.

src/regdesk/printer_manager.js
Cellivar: It lives on in a slightly different form. We need a method to, on page load, reconnect to all of the printers and load them up so the user doesn't have to click all over the place if they accidentally refresh the page or something.

 I've got a TODO to make it a bit more clever-er and thus more reliable at coming back in a stable state, right now it's a race condition on what printer lands in what slot. We should fix that before we ship.

src/regdesk/regdesk.js
Cellivar: So as you found eventually there are, indeed, two 'steps' of connecting to a printer. The first is granting the page access to talk to the printer _at all_, a necessarily manual process, and the actual connection thereof.

If the permission has been given in the past we can skip that step and jump right into the connect operation.

Cellivar:
Okay so everything is wired back up and working at this point. I'm able to print to the printers with test labels using the interface as it functions in this PR.

chrome_5B1wARV61T

The first time you're connecting to a printer the interface looks like this. Yes, the list items in the dropdown are apparently still taking up space even though they're hidden, which is weird but I actually kind of dig it for reasons you'll see in a moment.

Clicking connect on this dialog opens up the Chrome device selection dialog, which looks like this.

chrome_k1wFo4q1qC

This is where you select the device you want to pick up. Unfortunately because of REASONS the Zebra printers are all over the map in how useful the name here is. In this screenshot you can see one of the best results: we see the serial number right in the dialog. You can also see one of the worst results: just the name of the printer model. Alas we have no control over what displays in this box.

Once you've selected the device it will be connected to and loaded up, and the dropdown looks like this instead.

chrome_x1GbXDi5fX

And this is why I kind of dig the dropdown not resizing: the precise location of options is stable. You can't click on the 'connect' button anymore, it's not displayed at all, but the space where it would be is stable. Much more difficult to accidentally click on something you didn't mean to!

As the TODOs will indicate the configure screen doesn't exist yet. The other bits work, you can disconnect the printer (which lets you swap which printer is in which 'slot'), print a test pattern, feed a blank label, and that's working presently.

Finally here's what it looks like at rest with both printers connected.

chrome_b36oRPFZpM

Let me know if you have any questions or other feedback :D

I'd appreciate some tips on how to test some of this stuff. I attempted to follow good hygiene around dependency injection and avoiding coupling, but mocking out USB devices and the like seems.. involved? Might just be that I'm unfamiliar with the field.

src/regdesk/printer_dropdown.js
```diff
@@ -0,0 +1,204 @@
...
+    elements.forEach((e) => {
+      e.classList.remove('visible');
+      e.classList.add('invisible');
+    });
+  }
+
+  #show(...elements) {
+    elements.forEach((e) => {
+      e.classList.remove('invisible');
+      e.classList.add('visible');
```
kobaj, line 185: Good stuff, nicely managed with css as it should be! :D

src/regdesk/printer_dropdown.js
```diff
@@ -0,0 +1,204 @@
...
+  async feedLabel() {
+    if (!this.#printer) {
+      console.error("No printer selected, can't feed a label.");
+    }
+
+    try {
+      await this.#printer.feed(1);
+    }
+    catch (error) {
+      console.error(`Failed to feed label: ${error}`);
```
kobaj: For the record, this will print `Failed to feed label: [Object object]` which isn't very helpful. I personally prefer not formatting the string, instead going with `console.error('Failed to feed label', error);`

Try it yourself, paste these into your console:
```
console.error(`Failed to feed label: ${{error: { code: 400, message: 'heck'}}}`); // bad
console.error('Failed to feed label: %s', {error: { code: 400, message: 'heck'}}); // bad
console.error('Failed to feed label', {error: { code: 400, message: 'heck'}}); // good
```

src/regdesk/printer_dropdown.js
```diff
@@ -0,0 +1,204 @@
+import { LP2844 } from 'WebZLP/src/LP2844'
+
+export class PrinterDropdown extends EventTarget {
```
kobaj, line 3: Having this all be in a class rather than maintaining arrays is fantastic!

src/regdesk/regdesk.js
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * @param {Object} o - Argument container
+ * @param {PrinterManager} o.mgr - The printer manager to configure
+ * @param {PrinterDropdown} o.clearDropdown - The dropdown object managing the Clear printer.
+ * @param {PrinterDropdown} o.minorDropdown - The dropdown object managing the Minor printer.
+ * @param {Navigator} o.nav - The browser Navigator
+ */
+function configureManager({
+  mgr,
+  clearDropdown = new PrinterDropdown(PrinterDropdown.getElementsOnPage("adult", document)),
```
kobaj: I think you may have meant adultDropdown?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+   * Initializes a new instance of the PrinterManager class.
+   *
+   * @param {Navigator} nav - Browser Navigator to hook up events to.
+   * @param {PrinterDropdown} adultDropdown - The dropdown object managing the adult printer.
+   * @param {PrinterDropdown} minorDropdown - The dropdown object managing the Minor printer.
+   */
+  constructor(nav, adultDropdown, minorDropdown) {
+    this.#adultDropdown = adultDropdown;
+    this.#adultDropdown.addEventListener(PrinterDropdown.ConnectEventName, async (e) => this.handleRequestConnect(e));
+    this.#adultDropdown.addEventListener(PrinterDropdown.DisconnectEventName, async (e) => this.handleRequestDisconnect(e));
```
kobaj: DisconnectEventName isn't defined on the class PrinterDropdown?

src/regdesk/printer_dropdown.js
```diff
@@ -0,0 +1,204 @@
...
+    // Safe to call this without an await because there won't be a printer yet.
+    // TODO: kinda makes me itchy though.
+    this.removePrinter();
+  }
+
+  #connectPrinter() {
+    let event = new CustomEvent(
+      this.constructor.ConnectEventName,
+      { target: this,  });
+    this.dispatchEvent(event);
```
kobaj, line 90: I get what you're doing now. Thank you for the super cool pictures and explanation for the whole thing.

Not going to lie, it still feels a little funny to me. But I know I am very bad at translating "the funny feeling" into "here is why that is and here is how to fix it." Plus with time crunches and the fact you've rewritten this once (and its /way/ better this time already), I am happy to just let this go as is. :)

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+    try {
+      device = await navigator.usb.requestDevice({
+        filters: [
+          { vendorId: this.#printerType.usbVendorId }
+        ]
+      });
+    } catch (e) {
+      // TODO: One of the exceptions we can catch here is "user clicked cancel."
+      // Figure out how to separate that out as 'not an actual exception'.
+      console.error("Couldn't add new printer: " + e);
```
kobaj: I should have included this as an option above, but it will also print `[Object object]`.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+
+  /**
+   *
+   * @param {Object} o - Event Object.
+   * @param {USBDevice} o.device - The USB device to connect to as a printer.
+   * @returns
+   */
+  async handleConnectPrinter({ device }) {
+    // If we already have a tracked printer we shouldn't try to add it.
+    if (this.#dropdowns.some((d) => d.printer?.device === device)) {
```
kobaj, line 80: I see why you're not a `_.get()` follower, with your optional chaining! Back in my day...

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+      return;
```
kobaj: You might add a log message here explaining the early bail out.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    this.pairPrinter(dropdown);
+  }
+
+  /**
+   *
+   * @param {Object} o - Event Object.
+   * @param {USBDevice} o.device - The USB device to connect to as a printer.
```
kobaj, line 75: I don't think this is correct?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+      dropdown.setPrinter(printer);
+    } else {
+      console.warn("Connected to a printer that we couldn't assign anywhere!");
+    }
+  }
+
+  /**
+   * Handle a recquest to add a printer to a dropdown.
+   * @param {CustomEvent} o - PrinterDropdown connect event
+   * @param {PrinterDropdown} o.target - The dropdown target.
```
kobaj, line 61: Too many params?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+   */
+  assignPrinter(printer, dropdown) {
+    // Find the first dropdown in order, hope it's the right one!
+    // TODO: Make this smarter and pull from a cache or something for assignment
+    // instead of just first past the post.
+    dropdown = dropdown || this.#dropdowns.find(d => !d.printer);
+    if (dropdown) {
+      dropdown.setPrinter(printer);
+    } else {
+      console.warn("Connected to a printer that we couldn't assign anywhere!");
```
kobaj, line 53: Should we disconnect this printer?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Disconnect a printer from the page.
+   * @param {Object} o - Container object
+   * @param {USBDevice} o.device - The device to disconnect.
+   * @returns
+   */
+  async handleDisconnectPrinter({ device }) {
+    let dropdown = this.#dropdowns.filter(d => d.printer?.device === device);
```
kobaj: Technically each time you do dropdown.printer.device its breaking law of demeter. But you kind of have to with the current setup and its not too many levels deep, so I'm not going to be a stickler.

https://en.wikipedia.org/wiki/Law_of_Demeter

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+      // TODO: Display something more useful.
+      console.error("Failed to connect to printer: " + e);
+    }
+
+    return printer;
+  }
+
+  /**
+   * Disconnect a printer from the page.
+   * @param {Object} o - Container object
```
kobaj, line 99: Too many params?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+      return;
+    }
+
+    await dropdown.removePrinter();
+  }
+
+  /**
+   * Add a new printer to the page and connect to it.
+   * @param {PrinterDropdown} dropdown - The dropdown to add the printer to.
+   * @returns {LP2844} connected printer.
```
kobaj: Too many params?

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   * Disconnect and reconnect to all printers. NOTE: May shuffle position of printers!
+   */
+  async refreshPrinters() {
+    this.#dropdowns.forEach(async d => {
+      await d.removePrinter();
+    })
+
+    navigator.usb.getDevices().then((devices) => {
+      devices.forEach(async d => {
```
kobaj: another forEach async

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    await this.#minorDropdown.printer.printLabel(label);
+  }
+
+  /**
+   * Disconnect and reconnect to all printers. NOTE: May shuffle position of printers!
+   */
+  async refreshPrinters() {
+    this.#dropdowns.forEach(async d => {
```
kobaj: forEach async doesn't do what you think it does. Its a very dangerous trap, be careful.

https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+      await d.removePrinter();
+    })
+
+    navigator.usb.getDevices().then((devices) => {
```
kobaj: Missing an await (or return) statement on this promise, this async method behavior is undefined otherwise.

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  /**
+   * Assign a printer to a dropdown
+   * @param {LP2844} printer
+   * @param {PrinterDropdown} dropdown
+   */
+  assignPrinter(printer, dropdown) {
+    // Find the first dropdown in order, hope it's the right one!
+    // TODO: Make this smarter and pull from a cache or something for assignment
```
kobaj: I agree with you. But I don't think we need a cache or anything for assignment! Rather, the problem exists because we have two different methods trying to resolve this situation. Specifically refreshPrinters separates the dropdown from the printer, and assignPrinter tries to join them back together. But I think if refreshPrinters is doing the separation, it should be the one to do the join.

```
assignPrinter(printer, dropdown) {
  dropdown.setPrinter(printer);
}

async refreshPrinters() {
  const dropdownByPrinterSerial = await Promise.all(this.#dropdowns.map(async (dropdown) => {
    const serial = dropdown.printer.serial;
    await dropdown.removePrinter();
    return {serial: d.dropdownName};
  });

  return navigator.usb.getDevices().then(async (devices) => {
    return Promise.all(devices.map(async device => {
      const printer = await this.handleConnectPrinter({ device });
      const dropdown = this.#dropdowns.find(dropdown => dropdown.dropdownName === dropdownByPrinterSerial[printer.serial]);
      this.assignPrinter(printer, dropdown);
    });
  });
}
```

(This also fixes a number of promise issues, see my comments below).

src/regdesk/printer_manager.js
```diff
@@ -0,0 +1,171 @@
...
+    this.#adultDropdown.addEventListener(PrinterDropdown.DisconnectEventName, async (e) => this.handleRequestDisconnect(e));
+
+    this.#minorDropdown = minorDropdown;
+    this.#minorDropdown.addEventListener(PrinterDropdown.ConnectEventName, async (e) => this.handleRequestConnect(e));
+    this.#minorDropdown.addEventListener(PrinterDropdown.DisconnectEventName, async (e) => this.handleRequestDisconnect(e));
+
+    // Convenience array
+    this.#dropdowns = [this.#adultDropdown, this.#minorDropdown];
+
+    nav.usb.addEventListener('connect', async (e) => this.handleConnectPrinter(e));
```
kobaj: We don't do anything with this printer, worse off it gets garbage collected before we can disconnect, should we be handling this?

src/regdesk/printer_manager.js
Cellivar: When null propagation operators came to C# a million developers sighed with relief.

src/regdesk/printer_manager.js
Cellivar: I think here yes we do. This was generally meant as a final catch of the weird situation that someone physically hooked up 3 printers to one terminal (a feat, as there aren't that many USB ports..) as a 'just in case' thing. I'll add a dispose.

src/regdesk/printer_manager.js
Cellivar: I waffled on this one back and forth, and it's why PrinterDropdown is an EventTarget. The idea was that communication to and from the class would be wrapped in events. It'd keep the encapsulation closer.

What I settled on is a single direction ability of the PrinterManager to poke around inside of PrinterDropdown, while PrinterDropdown does not have any knowledge of the PrinterManager above it. This lets the dependency chain remain clean (no circular dependency) while avoiding a lot more boilerplate than I've already written.

If it ends up being an issue (or becoming more prevalent than it already is) we can shuffle around some boundaries to better close things up.

src/regdesk/printer_manager.js
Cellivar: Parameter destructuring! I came across the jsDoc syntax in a third party project I was working on for work last week, apparently this is the right way to document [anonymous objects for parameter destructuring](https://stackoverflow.com/a/42898969/1309423).

src/regdesk/printer_manager.js
Cellivar: I'm uncertain how you mean? Or you mean for the return? I clarified what's returned either way.

src/regdesk/printer_manager.js
Cellivar: image

This is how it ends up showing up in VSCode, the type hints forward through.

src/regdesk/printer_manager.js
Cellivar: Same as the other message, this is how you describe parameter destructuring.

src/regdesk/printer_manager.js
Cellivar: Same as the other message, this is how you describe parameter destructuring.

src/regdesk/printer_manager.js
Cellivar: Vestigial code, removed!

src/regdesk/regdesk.js
Cellivar: I do, that thing got renamed like 7 times and this got missed. Thanks!

src/regdesk/printer_manager.js
kobaj: oh interesting, cool!

src/regdesk/printer_manager.js
kobaj: Huh, that will take a while for me to get used to. Sorry for so many comments on it.

Unrelated: how are you embedding those pictures?

src/regdesk/printer_dropdown.js
Cellivar: Moved to your suggestion for all errors and warns that are sending an object.

src/regdesk/printer_manager.js
Cellivar: The object lifetimes for USBDevices are interesting as it will keep handing you the same instance of the same object.

When you call, for example, navigator.usb.requestDevice while you're already using a device somewhere else you won't be handed a _new_ USBDevice object, instead it'll be the same instance of the one you're using elsewhere. Closing the 'new' one we're handed would close the connection out from under the other printer.

At some point I'd like to make some utility wrapper or similar to PrinterManager in my library so it can better manage object lifecycles more directly because this surprised me, so it'll surprise others. I'm not able to find a great citation for this information, it was just handed to me by Some Guy I Know Who For Some Reason Knows A Lot About This Stuff ;)

Anyway, any situation in which we're handed a device object we should only dispose of it if we know we're the sole users of the object. This is fine for when we're handed a third printer when two are already hooked up, it's not owned by someone else and therefore we can dispose of it safely. In other cases though we aren't the sole owners and if anything we were handed the device by mistake and shouldn't do anything other than drop it on the floor as it's not our instance to mess with.

src/regdesk/printer_manager.js
kobaj: Woops, that first promise.all returns an array, not an object map. But I'm sure you figured that out. :)

src/regdesk/printer_manager.js
Cellivar: Drag and drop into the comment markdown field and it'll upload 'em.

I also use ShareX on my Windows machine for screenshot markup, that's how I do the boxes and things.

kobaj (dismissed):
The cleanup looks really good! Glad you were able to parse some of the exceptions too so they aren't as noisy for standard/expected operations. 

We are still talking past each other, I think, when it comes to the reconnect flow. But we can always tackle it later.

kobaj (approved):

PR #6, opened 2021-12-16
Add GHA validation workflow for PRs

Cellivar:
Should help us avoid surprises, especially as you add more tests to things.

This moves the mocha config over to a YAML file so that the test explorer extension I'm using works properly. It needed the config to be in a separate file as it doesn't read the package.json file.

image

If you're curious [this is the extension I'm using](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-mocha-test-adapter)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,6 +2,9 @@ name: Build
 
 on:
   pull_request:
+  push:
+    branches:
+      - main
 
 jobs:
   build:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,9 +18,10 @@ jobs:
       - run: npm ci --ignore-scripts
       - run: npm run test -- --reporter json --reporter-options output=test-results.json
       - name: Test Reporter
-        if: always()
+        if: success() || failure()
         uses: dorny/test-reporter@v1.5.0
         with:
+          name: 'Mocha Results'
           path: 'test-results.json'
           reporter: 'mocha-json'
           fail-on-error: 'true'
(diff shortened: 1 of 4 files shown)


Cellivar:
[Here's what a failed workflow ends up looking like](https://github.com/anthroarts/RegExtension/runs/4546612698)

Cellivar:
This will light up the little checkmarks next to the commitishes that float around on various screens. Keep an eye out for them and you can track what happened over time (and possibly where the first issue was encountered!).

Click on the details link to jump to the report for what happened. Note that the test report and the actual build are separate checks that display separate info.

image

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
kobaj, line 11: Interesting! I'm almost disappointed in Github with how limited configuration is allowed here. Gitlab allows you to specify a full docker name.

kobaj (approved):
Ship It!

kobaj:
Also holy cow you're amazing for including all those helpful screenshots and test commits! Thank you!

.github/workflows/build.yml
Cellivar: If we wanted to we could run the whole shebang in a container of our choice. The runs-on command controls which VM type it runs on. At work we have like 4 for various configurations.

_Within_ that VM you can then control what the job actually executes inside of. Here it'll run on the bare machine, but [we can run it in a container if we want to](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontainer) instead. The workflow syntax is fairly powerful in my experience, especially when you start laying into the conditional step systems.

PR #7, opened 2021-12-17
Feature/kobaj unit test all the things

kobaj:
More specifically I wrote unit tests for disable_auto_logout, popup, and event_page. In doing so I also cleaned up the mess I left behind earlier (ie deleted copy paste code, consolidated identifiers, and ensured execution only ran in the right stage).


src/disable_auto_logout.js
```diff
@@ -1,23 +1,32 @@
...
-  const intervalID = setInterval(() => {
+  setInterval(() => {
     if (!isEnabled) {
       return;
     }
 
+    // There is probably a better way to do this, utilizing something like
+    // angular.element($('[ng-controller]')).scope().<something>.clearIdle()
+    // But I couldn't figure it out. And hey, this works!
     window.dispatchEvent(new KeyboardEvent('keyup', null));
```
Cellivar, line 23: Maybe worth adding a comment on what this is specifically attempting to accomplish? When we're on the border of reverse engineering like this I find it useful to figure out what something was attempting to do when it inevitably breaks because of changes to the thing being reverse engineered.

Cellivar (dismissed):
LGTM! One nit about commenting, rest looks good. Nice work on the tests!

src/disable_auto_logout.js
kobaj: Do you mean what the `angular.element` is trying to accomplish? Or what I'm trying to accomplish with the `KeyboardEvent`?

Cellivar: The latter. While I generally understand that this is intended to prevent logouts and make the page happy, I'm uncertain why `keyup` is the mechanism to do so. Where did that information come from? You've documented a possible better way which is also helpful.

src/disable_auto_logout.js
```diff
@@ -9,9 +9,17 @@ const init = () => {
...
+    // the idle timers and if not called, will lead to the auto-logout popup. 
+    // 
+    // Ideally we would call this `clearIdle` directly, but its obfuscated 
+    // behind Angular. Theoretically we could grab a reference via something like
+    // `angular.element($('[ng-controller]')).scope().<something>.clearIdle()`
+    // But I couldn't figure it out.
+    //
+    // However, inside regfox is a window.addEventListener that triggers on 
+    // keyup, mousemove, and a few other window.Events and calls `clearIdle` 
+    // for us! So while a bit roundabout, this does reset idle timers too.
```
Cellivar, line 22: A+!

Cellivar (approved):

PR #8, opened 2021-12-17
Feature/kobaj regfox token refresh

kobaj:
This includes a few different commits
* Moved the original getBearerToken to regfox_page, since it isn't technically an API and is more like web scraping.
* Added an exchangeBearerToken method
  * Note, this does consume the old token, making it no longer usable
* Added a login method
  * Note, 2fa will be weird, I don't think I'm going to bother tackling it
  * The bearerToken returned from login is immediately usable
* Moved the queries to their own subfolder


Cellivar:
> Note, 2fa will be weird, I don't think I'm going to bother tackling it

At one point I left a tab open for a few hours and came back to a _javascript popup_ that was asking me to re-enter my password and 2FA token, which is bizzare but maybe doing some kind of shenanigans that we can also use. Up to you on how much fun you want to have chasing that down.

Cellivar (approved):
Woohoo shenanigans!

PR #9, opened 2021-12-18
Fix bug where some tabs wouldn't get messages.

kobaj:
Turns out you can have /two/ tabs open at the same time! Wow, technology these days is crazy.


Cellivar (approved):
Solved nicely!

PR #10, opened 2021-12-21
Add canvas-based label rendering

Cellivar:
This adds the class which takes the text content of the label and renders it to a canvas, which can then be handed to the printer for printing. The logic for that is wrapped up in the BadgeLabelBuilder class which handles the business of figuring this out.

I'm not completely happy with the way the offsets are laid out, and there's likely a future where that could be better configured via a JSON blob or something. For making our deadline I think hardcoding is acceptable.

This also adds the emoji font that we'll be using, which [comes from this repo](https://github.com/eosrei/twemoji-color-font/) and was converted to a WOFF2 file. The license for the font lets us use it with attribution which has been added to the README (and we should make sure it winds up in the extension folder as well somewhere).

The canvas element in the HTML file is a placeholder for debugging, demonstration, and testing purposes. Eventually the pre-print validation screen will show a canvas element like this for approval by the cashier.


public/regdesk.html
```diff
@@ -53,13 +53,19 @@
...
-        <div class="d-grid gap-2 mb-3">
-          <button class="btn btn-primary" id="newSearch">🔎 New Search</button>
+        <div class="row">
+          <div class="col-4 offset-4 d-grid gap-2 mb-3">
+            <button class="btn btn-primary" id="newSearch">🔎 New Search</button>
+          </div>
+        </div>
+        <div id="searchResults" class="row">
+          <!-- Demonstration canvas, actual layout will be different. -->
+          <canvas id="canvas" width="589" height="193" style="background-color: white; width: 589px; height: 193px; padding: 0;"></canvas>
```
kobaj, line 68: Double declaring the width/height?

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+  constructor (
+    alignmentPoint,
+    isBold,
+    maxWidth,
+    maxHeight,
+    offsetX,
+    offsetY
+  ) {
+    if (!["center", "left", "right"].includes(alignmentPoint)) {
+      throw new Error(`Invalid alignment point ${alignmentPoint}.`);
```
kobaj: This error message could be a little bit more helpful to users. Try to always give them something actionable when you throw a new exception. So for example.

```
const legalAlignmentPoints = ["center", "left", "right"];
if(!legalAlignmentPoints.includes(alignmentPoint)) {
  throw new Error(`invalid alignment point ${alignmentPoint}, expected one of ${legalAlignmentPoints}.`);
}
```

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
+import { LabelEpl } from 'WebZLP/src/LabelEpl'
+
+/**
+ * Line offset record
+ */
+class LineOffset {
```
kobaj, line 6: This is a fantastic "trait" class, thank you for making it! Its a perfect way to keep data and logic separated. (Think model in model - view - controller).

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Get the line offset information for lines on a badge.
+   */
+  static get lineOffsets() {
+    return {
+      // Manually calculated offsets to figure out a good mapping for a label
+      // with a width of 589px and height of 193px.
+      // These will absolutely need to be recalculated for other sizes.
+      // maxX of -1 means the label width
```
kobaj: Could be made a static global variable, that way later on when you do `if(offsets.maxX > -1)` the comment doesn't need to be duplicated. (So `if(offsets.maxX > LABEL_WIDTH)`)

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+      // offsetX of -1 means center it
+      "line1": new LineOffset("center", true, -1, 110, -1, 50),
```
kobaj: Arg comments might make it a little easier to see whats going on here.

```
new LineOffset("center", /* isBold= */ true, /* maxWidth= */ -1, /* maxHeight= */ 110, /* offsetX= */ -1, /* offsetY= */ 50);
```

In fact typing that out made me realize, whats -1 for offsetX mean?

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+
+    // TODO: Figure out a better way to do this?
+    let font = new FontFace("twemoji", 'url(fonts/TwitterColorEmoji-SVGinOT.woff2)', {
```
kobaj: const :P

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+    });
+    doc.fonts.add(await font.load());
+  }
+
+  /**
+   * Add this builder's details to a label.
+   * @param {LabelEpl} label - The label to add the print details to.
+   * @returns {LabelEpl} Label with the badge content added.
+   */
+  addToLabel(label) {
```
kobaj: Out variables and mutating variables considered /very/ bad. https://softwareengineering.stackexchange.com/questions/245767/is-modifying-an-incoming-parameter-an-antipattern

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+
+  /**
+   * Render this label's text onto a canvas, returning the ImageData.
+   * @param {number} width - The width of the label.
+   * @param {number} height - The height of the label.
+   * @returns {ImageData} - The ImageData of the canvas with the label's text rendered.
+   */
+  renderToImageData(width, height, canvas) {
+    canvas = canvas || new OffscreenCanvas(width, height);
+    const ctx = canvas.getContext('2d');
```
kobaj, line 148: Do you want to blank the canvas first just in case it has something else on it?

```
ctx.fillStyle = "white";
ctx.fillRect(0, 0, width, height);
```

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+    let fontSize = ctx.measureText(text);
+    let fontWidth = maxWidth / fontSize.width;
+
+    // Ascent and Descent are offsets from the 'centerpoint' of a character. Combined
+    // they represent the total space taken up by characters in the string.
+
+    // Note that the ascent and descent are calculated based on the character _set_
+    // that is used. If you have any lowercase characters it will calculate spacing
+    // for a lowercase 'g' even if there isn't one. No lowercase characters = No
+    // descenders calculated. This means allcaps is bigger for a given hegith limit.
```
kobaj: height*

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,197 @@
...
+   * @param {number} maxWidth - The max width of the bounding box, in pixels.
+   * @param {number} maxHeight - The max height of the bounding box, in pixels.
+   * @param {boolean} isBold - Whether the text should be bold.
+   * @returns The maximum font size to fit within the bounds of the label, in px.
+   */
+  #getMaxFontSizeInPx(ctx, text, maxWidth, maxHeight, isBold) {
+    // Rendering a font at 1px generates a multiplier to use for sizing.
+    // The offset will be dependent on the actual printed characters, so we
+    // need the real text that will be printed.
+    ctx.font = `${isBold ? "bold" : ""} 1px ${this.constructor.fontFamily}`;
```
kobaj: You copy paste this, consider putting it in a re-usable method?

src/regdesk/regdesk.js
```diff
@@ -25,10 +26,42 @@ function configureManager({
...
+    //   line1: "A Furry 🍑",
+    //   line2: "Just some furry name",
+    //   badgeId: "12345678",
+    //   level: "Super Sponsor",
+    //   isMinor: false,
+    // });
+
+    //let plabel = printerMgr.getAdultLabel();
+    //await label.addToLabel(plabel);
+    //await printerMgr.printAdultLabel(plabel);
```
kobaj: Would expect something much simpler...

```
printerMgr.printAdultLabel(label);
```

src/regdesk/regdesk.js
```diff
@@ -25,10 +26,42 @@ function configureManager({
...
+      badgeId: "12345678",
+      level: "Super Sponsor",
+      isMinor: false,
+    });
+
+    // You wouldn't normally call this directly, and instead use addToLabel like below
+    // Present here for demo purposes.
+    label.renderToImageData(canvas.width, canvas.height, canvas);
+
+    // Demo of the more average use, commented out so I stop accidentally printing labels.
```
kobaj, line 53: I super appreciate the demo code :D!

public/regdesk.html
Cellivar: Yes because something somewhere in Bootstrap was messing with it. Setting the value with either CSS or with the HTML values was insufficient, both were necessary in my (admittedly insufficient) testing. I planned to narrow down exactly what was going on there when I built out the proper interface instead of just this testing element.

src/regdesk/label_builder.js
Cellivar: C# calls 'em records, hence the jsdoc comment. I love using them for tiny bundles of state that get passed around, especially if I can make them immutable.

kobaj: But if you're aware of the awesome power that comes from immutable records, why isn't more of your code following that :P?

src/regdesk/label_builder.js
Cellivar: In this case no. Option 1 is we weren't handed a canvas element in which case we just created a blank one anyway. Option 2 is someone handed us a canvas to add things to, in which case we might assume they know what they want and we shouldn't surprise them by nuking their own canvas.

Disregarding "someone" is "me in one week when I forget half of this".

src/regdesk/label_builder.js
Cellivar: Good call, they differed very slightly so it didn't trip my mental code clone filter.

src/regdesk/label_builder.js
Cellivar: I have a vague memory in the back of my mind of some type of magical JS concept that was a type of flag, equal only to itself. I can't for the life of me find it right now. It'd be useful here instead of the more verbose enum pattern.

Looks like I can't define const public fields so getters it is.

src/regdesk/label_builder.js
Cellivar: Reasonable, with the printLabelBuilder method tweaks the return doesn't matter anymore so I've removed it.

In C# I'd [add an extension method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) for the LabelElp object to keep the dependencies nicely isolated. JavaScript [seems to have something similar](https://stackoverflow.com/a/9354310) but it makes me.. itchy. Would that be appropriate here?

src/regdesk/label_builder.js
Cellivar: I'm itchy about static globals because there doesn't appear to be a good way to make them immutable. Went with a static getter instead.

src/regdesk/label_builder.js
Cellivar: I didn't add the comments as I specifically wanted to keep the line from getting too long, expecting the IDE anyone is using to help guide them. I could switch it to using a parameter object with naming instead.

I've always found inline comments to be difficult to parse so I never think to try them.

src/regdesk/regdesk.js
Cellivar: Initially I wanted to avoid the extra dependency between the printer manager and the label builder but I don't know that it matters all that much. Added a method that does it all in one go.

src/regdesk/label_builder.js
Cellivar: I wrote the above comment in the context that C# [literally doesn't have the concept of a non-class-scoped field](https://stackoverflow.com/a/14368172) unlike JavaScript which does, in fact, have free-floating variables. The thought didn't even cross my mind to consider them until I reviewed your PR.

Would you rather I switch to them?

src/regdesk/printer_manager.js
```diff
@@ -184,6 +185,36 @@ export class PrinterManager {
...
   }
 
+  /**
+   * Print a label buider to the printer determined by the label builder.
+   * @param {BadgeLabelBuilder} builder
+   */
+  async printLabelBuilder(builder) {
+    var label;
+    switch (builder.isMinor) {
+      case true:
```
kobaj: I'm not a huge fan of switch/cases. I assume you're trying to avoid Javascript's falsy/truthy types, but you could just do a full equivalency check instead.

```
if (builder.isMinor === true) {

} else if (builder.isMinor === false) {

} else {
  // should log something
  return;
}
```

On the other hand, your switch/case did make me stop and go "wait a second, whats going on here" which was actually a good thing. So I could understand if you want to keep using it.

src/regdesk/label_builder.js
kobaj: Parameter object, builder pattern, etc might be better. But this also isn't going to be modified a lot or used by people. So we can just let it go through. :)

src/regdesk/label_builder.js
kobaj: I like your static getters, those are great!

src/regdesk/label_builder.js
```diff
@@ -106,19 +117,19 @@ export class BadgeLabelBuilder {
...
+   * @param {number} widthPadding - Dots to divide btween either side of the label as padding.
    */
-  addToLabel(label) {
+  addToLabel(label, heightPadding = 10, widthPadding = 20) {
     // Safety buffer around the edges of the printable area.
     // Make sure the width ends up on an byte boundary!
-    let widthSafeZone = 20;
-    let widthRemainder = (label.labelWidthDots - widthSafeZone) % 8;
-    let widthOffset = widthSafeZone + widthRemainder;
+    let widthRemainder = (label.labelWidthDots - widthPadding) % 8;
```
kobaj: Missed all these being const too :P

src/regdesk/label_builder.js
```diff
@@ -0,0 +1,217 @@
...
+  addToLabel(label, heightPadding = 10, widthPadding = 20) {
+    // Safety buffer around the edges of the printable area.
+    // Make sure the width ends up on an byte boundary!
+    let widthRemainder = (label.labelWidthDots - widthPadding) % 8;
+    let widthOffset = widthPadding + widthRemainder;
+    let canvasData = this.renderToImageData(
+      label.labelWidthDots - widthOffset,
+      label.labelHeightDots - heightPadding);
+
+    label.setOffset(widthOffset / 2).addImage(canvasData);
```
kobaj: This is still mutating the input variables, isn't it? Why can't we just avoid passing in the label all together?

```
getCanvasData(heightDots, widthDots, heightPadding = 10, widthPadding = 20) {
  const widthRemainder = ...;
  const widthOffset = ...;
  const canvasData = this.renderToImageData(
    widthDots - widthOffset,
    heightDots - heightPadding); 

 return {canvasData, widthOffset: widthOffset / 2};
}
```

src/regdesk/regdesk.js
```diff
@@ -25,10 +26,40 @@ function configureManager({
 }
 
 var printerMgr;
+var fontLoader = BadgeLabelBuilder.loadCustomFonts(document);
 
 document.addEventListener('readystatechange', async () => {
   if (document.readyState === "complete") {
     printerMgr = configureManager();
     await printerMgr.refreshPrinters();
+    await fontLoader;
```
kobaj, line 35: Is this doing anything?

src/regdesk/label_builder.js
Cellivar: Generally I've been playing it loose with `const` vs `let`, where the former is for when something really shouldn't change and the latter for whenever.

Should I be preferring const-only? I've found conflicting internet opinions on which to use when.

src/regdesk/regdesk.js
Cellivar: From line 29, yes.

To use the emoji font we need it to be loaded and available prior to attempting to use it in the canvas. I experimented with a few ways of loading fonts and ensuring they were available to both the page canvas elements as well as OffscreenCanvas elements. Explicitly loading it in JS and adding it to the Document's font list was the only _reliable_ way I could find to ensure this all happened. Other methods (CSS, hidden elements, etc) all had tradeoffs (hidden elements didn't actually load the font, CSS didn't load it into background canvas). 

Thus, we await loading the font. I'm not really happy with this solution, but it's expedient.

src/regdesk/printer_manager.js
Cellivar: The mental pattern I jumped to here is that isMinor is "a badge state modifier" and, in my experience, these types of modifiers grow. I then automatically jumped to "okay well if there's going to be more in the future this is deserving of a switch/case to handle them all". [C# has some interesting semantics around the switch/case keyword](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression#case-guards) that makes it much more common instead of chains of if statements, which is why I reached for it first.

That said, it's unlikely this specific iteration of the code will survive very long for that to be valuable. I'll swap it around to ifs.

src/regdesk/label_builder.js
Cellivar: It is, I mentioned [here](https://github.com/anthroarts/RegExtension/pull/10#discussion_r775310971) that I'd write an extension method that would be a neat(-er) way of handling this. I'm uncertain if the JS equivalent of mucking with object prototypes would be an appropriate solution here.

Basically, add a method onto the LabelEpl class that takes in the LabelBuilder as a parameter and adds the contents of the LabelBuilder to the LabelEpl. The relationship would be far more clear in that case. 

As for this particular method, since the information we need to modify the label is also present within the label, I presumed that it'd be easier to just work with the label directly instead of expecting it to be managed externally. 

My spidey sense goes off when I tend to pick some bits of an object out, pass them to a function, take the results, and then put them back into the same object. If the state is so closely related, why not just hand the state off entirely? 

I can move the label modification out.

src/regdesk/regdesk.js
kobaj: I guess I'm just lost on it not really being shown as a method call. Can we add the `();` to make that clear?

src/regdesk/label_builder.js
kobaj: I'm a const-only guy, I'm really curious to know who might disagree with that?

src/regdesk/label_builder.js
kobaj: Your spidey senses are right. That definitely weird. But its the whole "read some, write some" that's wrong in the first place. That is, an object shouldn't be used to mutate itself. Get into all kinds of hard to test/debug scenarios with that.

So you're right, my ultimate objection is refactor it /all/ to not do that. This is, in my mind, the first step. Decouple the Label from the LabelBuilder.

Cellivar: I've swapped it around and I do kinda feel that the complexity is now smeared across to the call site. I'm uncertain if the tradeoff there is worth it.

Really jonesing for my extension methods right now, they're one of the best ways to handle this kind of glue between classes..

src/regdesk/label_builder.js
Cellivar: [This popped up](https://overreacted.io/on-let-vs-const/) with a decent rundown of the then-common arguments. Generally came back as a 'talk amongst yourselves'.

Most of the languages I've worked with only have a concept of a one-time-assignment operation in terms of public object surface, so [.NET languages](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/readonly#ref-readonly-return-example), C++, etc. generally don't have the concept of inline variables having conditional assignment controls. It's a new brain worm to ingest :3

I can come back in another PR and add eslint or something to enforce the const-only style.

src/regdesk/label_builder.js
Cellivar: Thanks GitHub for dropping the websocket and not showing me a reply..

src/regdesk/regdesk.js
Cellivar: This might be a difference in how JS does async vs how I learned it. Is a promise a function you call? I thought the immediate return type of an async method is a promise _object_ that you then await in some fashion, not execute.

image

kobaj: Oooh, you're right. Didn't realize fontLoader was a promise.

src/regdesk/label_builder.js
kobaj: I think this is a lot better. The label is entirely self contained inside printLabelBuilder method. LabelBuilder doesn't need to know anything about it. The client constructs a LabelBuilder, passes it to PrinterManager without any back and forth. No mutation of variables. I honestly think this is the best and a super clean solution!

kobaj (approved):

PR #11, opened 2021-12-21
Feature/kobaj regfox get registration

kobaj:
Little api lets us get the user's date of birth (and like, everything else). Hope the documentation is clear.

Also fixed some issues with Errors printing [Object object]. Those little buggers pop up so much, they even trip me sometimes. :)

### DIFF
--- a/src/regfox/regfox_api.js
+++ b/src/regfox/regfox_api.js
@@ -28,11 +28,11 @@ const searchRegistrations = async (term, bearerToken) => {
   }).then(response => response.json())
     .then(response => {
       if (get(response, 'data.response.errors')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       if (!get(response, 'data.response.success', false)) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       return response.data.response;
@@ -53,11 +53,11 @@ const exchangeBearerToken = async (bearerToken) => {
   }).then(response => response.json())
     .then(response => {
       if (get(response, 'errors')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       if (!get(response, 'token')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       return response;
@@ -81,11 +81,11 @@ const login = async (email, password) => {
   }).then(response => response.json())
     .then(response => {
       if (get(response, 'data.mutationResponse.errors')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       if (!get(response, 'data.mutationResponse.success', false)) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       return response.data.mutationResponse;
@@ -106,11 +106,11 @@ const getRegistrationInfo = async (id) => {
   }).then(response => response.json())
     .then(response => {
       if (get(response, 'error')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       if (!get(response, 'data')) {
-        throw new Error(response);
+        throw new Error(`Could not parse or process response, see details to fix here: ${JSON.stringify(response)}`);
       }
 
       return parseRegfoxGetRegistrationResponse(response.data);
--- a/src/regfox/regfox_api.js
+++ b/src/regfox/regfox_api.js
@@ -96,7 +96,7 @@ const login = async (email, password) => {
  * Returns a nicely parsed object containing registration info given an id.
  * Returns a promise Error if the network is down or parsing failed.
  * 
- * @param {*} id of a registration, note this is **not** the RegistrationId, its just the id
+ * @param {string} id of a registration, note this is **not** the RegistrationId, its just the id
  */
 const getRegistrationInfo = async (id) => {
   const fullUrl = getRegistrationInfoUrl(id);
--- a/src/regfox/regfox_api.js
+++ b/src/regfox/regfox_api.js
@@ -1,7 +1,9 @@
 import { get } from 'lodash-es';
 import { buildSearchRegistrationsBody } from './queries/regfox_graphql_search_registrations_query.js';
 import { buildAuthLoginMutationBody } from './queries/regfox_graphql_auth_login_mutation_query.js';
+import { parseRegfoxGetRegistrationResponse } from './responses/regfox_api_get_registration_info_parser.js';
 
+const REGFOX_GET_REGISTRATION_URL = 'https://api.webconnex.com/v1/cp/report/regfox.com/registrant/${id}';
 const REGFOX_EXCHANGE_TOKEN_URL = 'https://api.webconnex.com/auth/exchange-token';
 const REGFOX_GRAPHQL_URL = 'https://api.webconnex.com/apollo/graphql';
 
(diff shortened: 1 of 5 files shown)


src/regfox/regfox_api.js
```diff
@@ -79,15 +81,45 @@ const login = async (email, password) => {
...
+      return parseRegfoxGetRegistrationResponse(response.data);
+    });
+}
+
+// Only exported for testing, DONT CALL THIS METHOD!
+const getRegistrationInfoUrl = (id) => {
+  return REGFOX_GET_REGISTRATION_URL.replace('${id}', id);
+}
+
+export { getRegistrationInfo, searchRegistrations, exchangeBearerToken, login, REGFOX_GRAPHQL_URL, REGFOX_EXCHANGE_TOKEN_URL, getRegistrationInfoUrl };
```
Cellivar, line 125: Thoughts on my `export <definition>` syntax vs this auto-suggested block format? Not a concern, just curious if you have an opinion.

src/regfox/regfox_api.js
```diff
@@ -79,15 +81,45 @@ const login = async (email, password) => {
...
+ * 
+ * @param {*} id of a registration, note this is **not** the RegistrationId, its just the id
+ */
+const getRegistrationInfo = async (id) => {
+  const fullUrl = getRegistrationInfoUrl(id);
+  return fetch(fullUrl, {
+    method: 'GET',
+    headers: buildHeaders(undefined)
+  }).then(response => response.json())
+    .then(response => {
```
Cellivar, line 107: In C# the equivalent of `.then()` is the `.ContinueWith()` method. It's become an antipattern over time as everyone suggests using async/await operations instead of a bunch of chaining. There's some specific reasons why in C# related to how the threading model will end up handling those that are related, but it's primarily related to keeping the code neat.

Any thoughts on that here? Awaiting individual operations vs chaining them together?

src/regfox/regfox_api.js
```diff
@@ -79,15 +81,45 @@ const login = async (email, password) => {
...
     });
 }
 
-export { searchRegistrations, exchangeBearerToken, login, REGFOX_GRAPHQL_URL, REGFOX_EXCHANGE_TOKEN_URL };
\ No newline at end of file
+/**
+ * Returns a nicely parsed object containing registration info given an id.
+ * Returns a promise Error if the network is down or parsing failed.
+ * 
+ * @param {*} id of a registration, note this is **not** the RegistrationId, its just the id
```
Cellivar: Should this be a string?

```suggestion
 * @param {string} id of a registration, note this is **not** the RegistrationId, its just the id
```

test/regfox/responses/regfox_api_get_registration_info_parser.test.js
```diff
@@ -0,0 +1,72 @@
+import { expect } from 'chai';
+
+import { parseRegfoxGetRegistrationResponse } from '../../../src/regfox/responses/regfox_api_get_registration_info_parser.js';
+
+describe('regfox_api_get_registration_info_parser', () => {
+  describe('parseRegfoxGetRegistrationResponse', () => {
+    it('doesn\'nt crash and burn on undefined input', () => {
```
Cellivar: Personal nit: When writing messages like this I avoid contractions (don't -> do not) just to avoid the backslash in the middle of the word. Something about it trainwrecks my reading comprehension every time. Not a blocker, just a nitpick.

test/regfox/responses/regfox_api_get_registration_info_parser.test.js
```diff
@@ -0,0 +1,72 @@
...
+      const body = parseRegfoxGetRegistrationResponse({
+        id: 1234, registrationOptions: 'standard', data: [{
+          "path": "registrationOptions",
+          "value": "option1",
+        }]
+      });
+      expect(body).to.deep.include({ id: 1234, registrationOptions: "standard" });
+    });
+
+    it('gets the god damn birthdate like we wanted from the search api', () => {
```
Cellivar: https://twitter.com/gitlost

test/regfox/regfox_api.test.js
```diff
@@ -136,4 +154,36 @@ describe('regfox_api', () => {
...
+      const result = await getRegistrationInfo(123455);
+
+      expect(result).to.include({ 'transactionCount': 1 });
+    });
+
+    it('handles a failure in the API', async () => {
+      getResponse.returns({
+        error: ['heky!'],
+      });
+      return getRegistrationInfo(123455).should.eventually.be.rejected;
```
Cellivar, line 186: Something about super fluent APIs that makes me itchy over time. Reminds me of trying to write Cucumber syntax ages ago...

test/regfox/regfox_api.test.js
```diff
@@ -136,4 +154,36 @@ describe('regfox_api', () => {
...
+        "data": {
+          "id": 123455,
+          "accountId": 1311,
+          "registrationId": "593409403",
+          "transactionCount": 1
+        }
+      });
+      const result = await getRegistrationInfo(123455);
+
+      expect(result).to.include({ 'transactionCount': 1 });
```
Cellivar, line 179: Moot to check for other details?

Valuable to do any type checking (ID is a number vs string)?

src/regfox/responses/regfox_api_get_registration_info_parser.js
```diff
@@ -0,0 +1,329 @@
...
+    }
+    return { [i.path]: i.value, ...acc };
+  }, {});
+  return { ...dataObject, ...rawResponse };
+}
+
+export { parseRegfoxGetRegistrationResponse };
+
+// The raw response looks something like the below:
+/*{
```
Cellivar, line 36: Worth dropping into a .json file in the tests directory as "this is what a real API result looks like" ?

src/regfox/responses/regfox_api_get_registration_info_parser.js
```diff
@@ -0,0 +1,329 @@
...
+ * 
+ * @param {*} rawResponse see below
+ * @returns 
+ */
+const parseRegfoxGetRegistrationResponse = (rawResponse) => {
+  const dataObject = rawResponse?.data?.reduce((acc, i) => {
+    if (!i.value) {
+      return acc;
+    }
+    return { [i.path]: i.value, ...acc };
```
Cellivar, line 28: This took me a minute, clever!

Cellivar (dismissed):
Handful of non-blocking nits, actual code itself LGTM.

src/regfox/regfox_api.js
kobaj: I'm slightly partial towards the export being at the bottom for two reasons
1. That's how old javascript did it :P
2. Keeps all of the exports in one single location

However, if we are only exporting a single method/class from a file I'm happy to use `export <definition>` instead.

src/regfox/regfox_api.js
kobaj: Definitely never mix the two with something like `.then(async () => await...)`. Otherwise, I have no strong feelings one way or another.

In a case like this where we would have a bunch of extra variables laying about (`response`, `jsonResponse`, `jsonDataResponse`, etc) as we process things down to the final "response" I think `.then` is slightly nicer in that those intermediates don't get exposed unnecessarily.

src/regfox/regfox_api.js
kobaj: You and your...types. 👍

test/regfox/responses/regfox_api_get_registration_info_parser.test.js
kobaj: Totally fair! Will fix. :)

test/regfox/responses/regfox_api_get_registration_info_parser.test.js
kobaj: Yeah, that was a moment of unprofessionalism on my part. Will fix :).

test/regfox/regfox_api.test.js
kobaj: I definitely don't like how these aren't methods. It feels weird to `.object.object.object` instead of `.method().method().method()`. But having said that, I /looooove/ fluent APIs! Chai comes with [Asserts](https://www.chaijs.com/api/assert/) and I won't stop you if you want to use them. But I'm using Chai's fluent expectations personally :D.

test/regfox/regfox_api.test.js
kobaj: I'm happy to check for other properties being present in the response (admittedly I was fairly lax when checking here). Was there some particular ones you wanted to include?

But I don't think we should be /too/ tightly coupled that if they change a type (but semantically an id is still an id), our code fails. (Now, canary tests would be a different story, but I don't think we can set those up). Basically since javascript is pretty type lenient, I'm following those same practices.

src/regfox/responses/regfox_api_get_registration_info_parser.js
kobaj: Its not currently legal json. If there is a "foo" account I can just grab some fake data from, happy to do it yeah!

src/regfox/responses/regfox_api_get_registration_info_parser.js
kobaj: You should have seen the code I originally found on the internet.

```
return acc[i.path]: i.value, acc;
```
That one took me a very long time to realize whats going on, and rewrote it very quick once I did haha.

Cellivar (approved):
LGTM!

test/regfox/responses/regfox_api_get_registration_info_parser.test.js
Cellivar: I was more amused than scolding ;)

PR #12, opened 2021-12-28
Add ESLint, goog style, with some minor tweaks

Cellivar:
Made a judgement call based on the differences between major styleguides (Airbnb, Google, 'Standard') and chose the one that seemed to have the lease madness involved compared to how we've been writing code.

The following custom rules were changed:

* `'linebreak-style': 'off',` - Line endings don't matter here.
* `'max-len': 'off',` - I dislike hard max line lengths and instead prefer a soft target of 80.
* `'prefer-const': 'error',` - As previously discussed we want to prefer const.
* `'object-curly-spacing': ['error', 'always'],` - My autoformatter inserts these and they make it more legible.

For reasons I cannot fathom the default indent size of function parameters is different from the function block and that breaks my brain so we set it back to normal.
```'
indent': ['error', 2, {
  'FunctionDeclaration': {
    'parameters': 1,
    'body': 1,
  },
```


Cellivar:
I'm expecting a single linter failure in this build..

Cellivar:
Ah and [there it is](https://github.com/anthroarts/RegExtension/runs/4647101457?check_suite_focus=true#step:5:11). Not very pretty, at least legible enough.

src/content_script.js
```diff
@@ -1,74 +1,73 @@
-import $ from "jquery";
+/* eslint-disable require-jsdoc */
```
Cellivar, line 1: I figure this script isn't long for this world in this format anyway so documenting it by hand wouldn't be useful.

kobaj (approved):
Super solid! Glad you were able to add eslint. Its my favorite and I agree with all of your customizations over the default :).

PR #13, opened 2021-12-31
Add workflow screen state machine framework

Cellivar:
Okay so [I read a thing](https://dev.to/davidkpiano/you-don-t-need-a-library-for-state-machines-k7h) and pulled some ideas from it.

Our workflow for cashier processing can be broken down into a fairly straightforward workflow without a lot of looping and complex operations. I'm working on adding a _complete_ workflow diagram of the workflows we need to support. A complete workflow is something we can represent as a state machine.

So I implemented a state machine.

The basic idea is that each state class encapsulates the necessary stuff to wire up the screen to do whatever that step of the workflow is capable of doing. In this PR are three example classes that handle search, waiting for search results to return, and then displaying the result. The search to loading transition also demonstrates support for arbitrary data transfer across a transition boundary.

We handle the transitions via events to better ensure no circular dependency references. 

The entire state machine graph will be described in the state machine manager class. I'm not 100% on this solution and am open to ideas on a better representation.

I'm submitting this so I can go to bed.

### DIFF
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -13,7 +13,7 @@
 </head>
 
 <body>
-  <div class="container-fluid container-no-gutter">
+  <div class="container-fluid container-no-gutter container-fullheight-with-navbar">
     <nav class="navbar navbar-expand-lg navbar-dark bg-dark">
       <div class="container-fluid container-no-gutter d-flex">
         <span class="navbar-brand me-auto" href="#">
@@ -53,23 +53,14 @@
         </div>
       </div>
     </nav>
-    <div class="row row-no-margin container-no-gutter">
-      <div id="buttonGutterLeft" class="col-2 d-grid gap-2 container-no-gutter">
-        <button id="cancelButton" class="btn btn-danger btn-fullheight">CANCEL</button>
+    <div class="row row-no-margin container-no-gutter h-100">
+      <div id="buttonGutterLeft" class="col-2 d-grid gap-2 h-100 container-no-gutter-left">
+        <button id="leftButton" class="btn btn-danger btn-fullheight btn-huge">CANCEL</button>
       </div>
-      <div class="col-8 pt-3">
-        <div class="row">
-          <div class="col-4 offset-4 d-grid gap-2 mb-3">
-            <button class="btn btn-primary" id="newSearch">🔎 New Search</button>
-          </div>
-        </div>
-        <div id="searchResults" class="row">
-          <!-- Demonstration canvas, actual layout will be different. -->
-          <canvas id="canvas" width="589" height="193" style="background-color: white; width: 589px; height: 193px; padding: 0;"></canvas>
-        </div>
+      <div class="col-8 pt-3" id="centerField">
       </div>
-      <div id="buttonGutterRight" class="col-2 d-grid gap-2 container-no-gutter">
-        <button id="printButton" class="btn btn-success btn-fullheight">PRINT</button>
+      <div id="buttonGutterRight" class="col-2 d-grid gap-2 h-100 container-no-gutter-right">
+        <button id="rightButton" class="btn btn-success btn-fullheight btn-huge">PRINT</button>
       </div>
     </div>
   </div>
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -54,13 +54,38 @@
       </div>
     </nav>
     <div class="row row-no-margin container-no-gutter h-100">
-      <div id="buttonGutterLeft" class="col-2 d-grid gap-2 h-100 container-no-gutter-left">
-        <button id="leftButton" class="btn btn-danger btn-fullheight btn-huge">CANCEL</button>
+      <div class="col-2 d-grid gap-2 h-100 container-no-gutter-left">
+        <button id="cancelBtn" class="btn btn-danger btn-fullheight btn-huge">CANCEL</button>
       </div>
       <div class="col-8 pt-3" id="centerField">
+        <div id="newSearchState" class="row d-none">
+          <form class="row g-3" id="newSearchForm">
+            <div class="col-md-9">
+              <label for="newSearchText" class="form-label">Legal Name or Scan Barcode</label>
+              <input class="form-control form-control-lg" id="newSearchText" type="text" />
+            </div>
+            <div class="col-md-3 d-grid">
+              <button type="submit" class="btn btn-primary btn-lg" id="newSearchBtn">Search🔎</button>
+            </div>
+          </form>
+        </div>
+        <div id="loadingSearchResultState" class="row d-none">
+          <div class="col-md-8 offset-md-2"><h2>Searching for "<span id="loadingSearchText"></span>".</h2></div>
+          <div class="col-md-1">
+            <div class="spinner-border" role="status">
+              <span class="visually-hidden">Loading...</span>
+            </div>
+          </div>
+        </div>
+        <div id="singleResultState" class="row d-none">
+          <div id="searchResults" class="row">
+            <!-- Demonstration canvas, actual layout will be different. -->
+            <canvas id="canvas" width="589" height="193" style="background-color: white; width: 589px; height: 193px; padding: 0;"></canvas>
+          </div>
+        </div>
       </div>
-      <div id="buttonGutterRight" class="col-2 d-grid gap-2 h-100 container-no-gutter-right">
-        <button id="rightButton" class="btn btn-success btn-fullheight btn-huge">PRINT</button>
+      <div class="col-2 d-grid gap-2 h-100 container-no-gutter-right">
+        <button id="printBtn" class="btn btn-success btn-fullheight btn-huge">PRINT</button>
       </div>
     </div>
   </div>
--- a/src/regdesk/states/loading_state.js
+++ b/src/regdesk/states/loading_state.js
@@ -3,7 +3,7 @@ import { RegState } from './reg_state.js';
 /**
  * Loading state
  */
-export class LoadingState extends RegState {
+export class LoadingSearchResultState extends RegState {
   /**
    * Get the transition events this state can process.
    */
@@ -22,7 +22,9 @@ export class LoadingState extends RegState {
    * @param {*} eventMap - Mapping of this object's transition events to new states.
    */
   constructor(regMachineArgs, eventMap) {
-    super(regMachineArgs, eventMap);
+    super(regMachineArgs, eventMap, 'loadingSearchResultState');
+
+    this.searchValueSpan = this.screenRow.querySelector('#loadingSearchText');
   }
 
   /**
@@ -30,15 +32,16 @@ export class LoadingState extends RegState {
    * @param {CustomEvent} e - The event details object.
    */
   enterState(e) {
-    this.leftButton.show().setTransitionCallback(this, LoadingState.events.CANCEL);
-    this.rightButton.hide();
+    this.show(this.screenRow);
+    this.cancelButton.visible().setTransitionCallback(this, LoadingSearchResultState.events.CANCEL);
+    this.printButton.invisible();
+
+    this.searchValueSpan.innerText = e.detail.searchText;
 
     // Demo code!
-    this.centerField.innerText = `Searching for ${e.detail.searchText}`;
-
     // Actually wait for results from commMgr instead of this lol.
     setTimeout(() => {
-      this.dispatchTransition(LoadingState.events.SINGLE_RESULT_READY);
+      this.dispatchTransition(LoadingSearchResultState.events.SINGLE_RESULT_READY, { badgeLine1: e.detail.searchText });
     }, 1000);
   }
 
@@ -46,6 +49,9 @@ export class LoadingState extends RegState {
    * Exit this state.
    */
   exitState() {
-    this.leftButton.clearTransitionCallback();
+    this.hide(this.screenRow);
+    this.cancelButton.clearTransitionCallback();
+
+    // TODO: cancel any in-progress searches!
   }
 }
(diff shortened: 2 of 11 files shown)


public/regdesk.html
```diff
@@ -53,23 +53,14 @@
...
-        <div id="searchResults" class="row">
-          <!-- Demonstration canvas, actual layout will be different. -->
-          <canvas id="canvas" width="589" height="193" style="background-color: white; width: 589px; height: 193px; padding: 0;"></canvas>
-        </div>
+      <div class="col-8 pt-3" id="centerField">
       </div>
-      <div id="buttonGutterRight" class="col-2 d-grid gap-2 container-no-gutter">
-        <button id="printButton" class="btn btn-success btn-fullheight">PRINT</button>
+      <div id="buttonGutterRight" class="col-2 d-grid gap-2 h-100 container-no-gutter-right">
+        <button id="rightButton" class="btn btn-success btn-fullheight btn-huge">PRINT</button>
```
kobaj: I think this was more accurate when it was named printButton?

src/regdesk/states/loading_state.js
```diff
@@ -0,0 +1,51 @@
+import { RegState } from './reg_state.js';
+
+/**
+ * Loading state
+ */
+export class LoadingState extends RegState {
```
kobaj: One of these days you'll learn inheritance is evil and should be avoided at all costs.

public/regdesk.html
```diff
@@ -53,23 +53,14 @@
         </div>
       </div>
     </nav>
-    <div class="row row-no-margin container-no-gutter">
-      <div id="buttonGutterLeft" class="col-2 d-grid gap-2 container-no-gutter">
-        <button id="cancelButton" class="btn btn-danger btn-fullheight">CANCEL</button>
+    <div class="row row-no-margin container-no-gutter h-100">
+      <div id="buttonGutterLeft" class="col-2 d-grid gap-2 h-100 container-no-gutter-left">
+        <button id="leftButton" class="btn btn-danger btn-fullheight btn-huge">CANCEL</button>
```
kobaj: I think this was more accurate as cancelButton.

src/regdesk/states/reg_machine_args.js
```diff
@@ -0,0 +1,49 @@
...
+    commManager,
+  ) {
+    this.leftButton = leftButton;
+    this.rightButton = rightButton;
+    this.centerField = centerField;
+    this.printerManager = printerManager;
+    this.commManager = commManager;
+
+    if (this.constructor === RegMachineArgs.constructor) {
+      Object.freeze(this);
```
kobaj, line 29: :D! Nice! I love the immutability!

src/regdesk/states/loading_state.js
kobaj: I'm wondering if you're wanting to use this LoadingState in a general sense (like anytime there is a need to load) or only when specifically loading a result from, for example, search? If the former, I have concerns. If the latter, it should be named more specific :).

Edit: After reading the rest of the code it seems like this class is a mix, but slightly leaning more towards the latter.

src/regdesk/states/loading_state.js
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Actions to perform when entering the state.
+   * @param {CustomEvent} e - The event details object.
+   */
+  enterState(e) {
+    this.leftButton.show().setTransitionCallback(this, LoadingState.events.CANCEL);
+    this.rightButton.hide();
+
+    // Demo code!
+    this.centerField.innerText = `Searching for ${e.detail.searchText}`;
```
kobaj: Ideally you shouldn't set text via javascript. Makes i18n a nightmare. You would just show/hide elements from javascript and keep all the translations in the html.

src/regdesk/states/new_search_state.js
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Enter this state.
+   */
+  enterState() {
+    this.leftButton.hide();
+    this.rightButton.hide();
+
+    this.centerField.innerHTML = this.#getCenterTemplate();
+    this.centerField.querySelector('#newSearchText').focus();
```
kobaj: Feels kind of weird to have the RegState handling most of the dom retrieval (eg querying for left/right button) but then this class has to also do dom retrieval. Can they be centralized?

src/regdesk/states/new_search_state.js
```diff
@@ -0,0 +1,65 @@
...
+        const searchText = this.centerField.querySelector('#newSearchText').value;
+        this.dispatchTransition(NewSearchState.events.START_SEARCH, { searchText: searchText });
+      });
+  }
+
+  /**
+   *
+   * @return {string} - The center html elements.
+   */
+  #getCenterTemplate() {
```
kobaj: Kind of like translations, your javascript shouldn't really be injecting html either. Just hiding/showing the correct fields. That way all of your HTML is in one place, nicely separated from your javascript. And if you need to change one, you don't have to worry if you missed something in the other.

src/regdesk/states/reg_machine_manager.js
```diff
@@ -0,0 +1,106 @@
...
+    // TODO: Probably a way to merge the event map with the machine declaration
+    // whem I'm not mentally fried on my third flight of the day.
+    eventMap.forEach((state) => {
+      // Interesting, eslint doesn't handle this correctly.
+      // eslint-disable-next-line new-cap
+      machine.states[state.type.name] = new state.type(regMachineArgs, state.events);
+
+      for (const [eventName] of Object.entries(state.events)) {
+        machine.states[state.type.name]
+          .addEventListener(eventName, async (e) => await this.transition(e));
```
kobaj: Unnecessary await, surprised that eslint didn't catch that.

src/regdesk/states/reg_machine_manager.js
```diff
@@ -0,0 +1,106 @@
...
+      },
+      {
+        type: SingleResultState,
+        events: {
+          [SingleResultState.events.CANCEL]: NewSearchState.name,
+        },
+      },
+    ];
+
+    // TODO: Probably a way to merge the event map with the machine declaration
```
kobaj, line 60: Something like this I imagine?
```
const states = eventMap.reduce((acc, state) -> {
  const typedState = new state.type(regMachineArgs, state.events);
  // for loop here
  return {state: typedState, ...acc};
}, {});

const machine = {initial: NewSearchState.name, states}
```

src/regdesk/states/reg_machine_manager.js
```diff
@@ -0,0 +1,106 @@
...
+    } catch (e) {
+      // TODO: Display an error or something.
+      console.error(`Failed to run actions from ${state} exit event ${event.type}.`, e);
+    }
+
+    try {
+      this.#machine.states[nextStateName]?.enterState?.(event);
+    } catch (e) {
+      // TODO: Display an error or something.
+      console.error(`Failed to run actions from ${state} enter event ${event.type}.`, e);
```
kobaj, line 101: Excellent error messages :D!

src/regdesk/states/reg_machine_manager.js
```diff
@@ -0,0 +1,106 @@
...
+   * @param {RegMachineArgs} regMachineArgs - The standard record for each state class.
+   */
+  constructor(regMachineArgs) {
+    const machine = {
+      initial: NewSearchState.name,
+      states: {},
+    };
+
+    // TODO: Is there a better way to store this?
+    const eventMap = [
```
kobaj, line 37: Eh, ultimately not how I would have done things, but I'm not picky. She'll do for our needs for now I suspect.

src/regdesk/workflow_button.js
```diff
@@ -0,0 +1,100 @@
...
+    this.#transitionEvent = event;
+    this.#transitionEventDetails = details;
+    this.#element.addEventListener(this.#buttonClickEvent, this.handleClickEvent.bind(this));
+    return this;
+  }
+
+  /**
+   * Dispatch a click event, if relevant to do so.
+   *
+   * @description Okay but why though? So we can remove this event listener.
```
kobaj, line 79: Haha, agreed, feels a little funny.

kobaj (commented):
Most of my comments are nitpicky things. The general concept of using a state machine irks me, but I don't mind if you want to try it and experiment. As such your implementation seems good. I think this is ripe for some unit testing though!

src/regdesk/states/reg_machine_args.js
Cellivar: I stumbled across it when looking at some library that had a whole bunch of different ways of creating / generating various types of immutable things. The constructor guard makes sure it's _this_ constructor we called, as opposed to an extending class which should be doing its own freezing in its own constructor.

src/regdesk/states/loading_state.js
Cellivar: This is pretty much the only time I ever use it anymore: there's a handful of behaviors that everything needs to do the same and everything else is implementation specific.

Most of my C# these days is all classes with a batch of interfaces, but JS doesn't have interfaces.

Proper polymorphism-based shenanigans have come back to haunt me too many times to go looking to implement them anymore.

src/regdesk/workflow_button.js
Cellivar: Man I spent so much time trying to find a better system for re-using UI elements with different callbacks.

src/regdesk/states/loading_state.js
Cellivar: Correct, I expect this to end up as a dedicated state in between specifically performing a search for an identity and showing the results of those identities.

There's two other possible locations for a loading state: Hitting the API to mark someone paid, and hitting the API to check someone in. The paid status will likely end up being an aspect of the single result page anyway and not a dedicated state. Check-in will happen alongside badge print, and will likely also not be a separate state. Still hammering out details into the state diagram I'm building.

src/regdesk/states/loading_state.js
Cellivar: I agree, hence the above note that this is demo code. Mercifully isn't a feature (for _this_ con anyway) but is something I attempt to keep in mind either way.

src/regdesk/states/new_search_state.js
Cellivar: This specific line was a bit of an afterthought to see how hard it would be to ensure that when the page arrives it's ready to be typed into. 

That said, I consider the page-state-object to be _scoped_ to controlling the center field while it's entered the state. The enter method should be doing whatever is necessary to make that center field ready to go for the state's behavior which may include performing DOM queries within the field it's controlling. I don't think centralizing the painting of the center field is going to work well when each state object is going to be doing its own thing with it.

There's a split here between painting the page and handling the behaviors of the page that, given more time for a more robust architecture, I'd prefer to build. I just don't know that I have time..

src/regdesk/states/new_search_state.js
Cellivar: Ahha this brings the previous comment into better focus. Let me see if I can migrate stuff over.

src/regdesk/states/reg_machine_manager.js
Cellivar: Vestigial from a previous iteration, I'll do a pass to see what else might need cleanup.

Cellivar:
> The general concept of using a state machine irks me, but I don't mind if you want to try it and experiment.

Definitely open to hearing alternate ideas. With the fairly straightforward workflow with limited looping a state machine was the first thing I considered. My VHDL days sometimes influence how I consider problems, so I'm always happy to hear how more CS-leaning folks think about it!

> I think this is ripe for some unit testing though!

Goodness yes they are. I wanted to get the overall design in front of you before I circled back on them.

public/regdesk.html
Cellivar: Good point, rolled it back but with "btn" to match the CSS selector convention.

public/regdesk.html
Cellivar: Good point, rolled it back but with "btn" to match the CSS selector convention.

kobaj:
> 

In larger scale systems I would push for more of a decision table like approach. That is, rather than [State A] leads to [State B]. There is just a set of axioms `isPaid`, `isPrinted`, etc. And the [State A] is determined when `isPaid = false && IsPrinted = false`. Whereas [State B] is determined when `isPaid = true && isPrinted = false`.

https://en.wikipedia.org/wiki/Decision_table

My bias stems from the standpoint of I've never seen a state machine that I liked, haha. But you may have convinced me here. 

Especially since 
1. The machine should be doing the exact same steps a human would (and humans think really similarly to state machines)
1. The logic is very simple, and has a hard limit on its ability to grow (IE, we only have ~9 days to finish code)
1. All of this is meant to "guide" the human step by step anyway (arguably the same reason as 1, but whatever)

src/regdesk/states/loading_state.js
Cellivar: Clarified class name.

src/regdesk/states/new_search_state.js
Cellivar: Okay in response to the three related comments, each state manages its own dedicated _row_ instead of the entire center element. The row IDs are derived directly off the class name automatically and are the only thing the RegState class exposes to the derived classes.

This moves the HTML over to just the HTML and we limit selector operations to just the class constructors.

Individual state classes still need to do their own dom queries unless we want to have one giant map object which I'm adamantly against based on previous horror stories involving autogenerated object maps for automated web GUI testing. I think scoped queries are okay here as we've given the state classes a sandbox within which they're meant to play.

src/regdesk/states/new_search_state.js
Cellivar: See above, no more injection.

src/regdesk/states/loading_state.js
Cellivar: See below, moved to a separate span element within a static html string.

Cellivar:
Turning on auto-merge in case you get to this in the morning before I do.

kobaj (approved):

PR #15, opened 2022-01-01
Add remaining workflow states and events

Cellivar:
Smol PR, after building the workflow in the document I went back in and reviewed the gaps. I think this is it?

There's an argument to be made that unpaid vs paid is a separate state that should be explicitly defined, right now it's a detail constrained to the single result class, which has an event to refresh itself (re-pull details from the API) and reload the page. 

Next up: Tests, then probably the commMgr class so we can start making API calls.

Cancel button is wired up and works as expected now for passing through multiple results. If you get two results, click one of them, then cancel, you'll end up back at the list of multiple results.

Oh, and it turns out that removing event listeners is even more involved than I thought and the way I was trying to do it was still insufficient. I've given up on that entirely now.


kobaj:
> There's an argument to be made that unpaid vs paid is a separate state that should be explicitly defined, right now it's a detail constrained to the single result class, which has an event to refresh itself (re-pull details from the API) and reload the page.

I think you're starting to hit one of the fundamental questions for state machines. When is a new state necessary vs augmenting an existing one. I think you need to be very clear in definition for this. Especially because the same question could apply to single_result_ready vs multiple_result_ready.

src/regdesk/states/multiple_result_state.js
```diff
@@ -0,0 +1,55 @@
+import { RegState } from './reg_state.js';
+
+/**
+ * Multiple result state
+ */
+export class MultipleResultState extends RegState {
+  /**
+   * Get the transition events this state can process.
+   */
+  static get events() {
```
kobaj, line 10: I am just now realizing you define these events a state is responsible for twice. Once inside the state and once inside the mgr. Can you reduce that down to just one? I think having it contained in just the state is best.

src/regdesk/states/multiple_result_state.js
```diff
@@ -0,0 +1,55 @@
...
+   * @param {*} eventMap - Mapping of this object's transition events to new states.
+   */
+  constructor(regMachineArgs, eventMap) {
+    super(regMachineArgs, eventMap);
+
+    this.button1 = this.screenRow.querySelector('#multipleResultExampleOne');
+    this.button1.addEventListener('click', () => {
+      this.dispatchTransition(
+        MultipleResultState.events.SELECTED_SINGLE_RESULT,
+        { badgeLine1: 'ASDF', fromMultiple: true });
```
kobaj: The "fromMultiple" should not be part of the response details. Mostly because it's the job of the mgr to know the flow of states and this is kind of going around that. I would expect mgr to be smart enough to know a cancel from here leads back to multiple vs single. Otherwise you're kind of blurring the boundaries between the responsibilities of a state (present information and buttons) and the mgr (handle responses to those buttons ie state transitions).

src/regdesk/states/multiple_result_state.js
Cellivar: This static getter here is where we define the events that a state can handle, while what we define in the mgr class is the relationship between a state's events and the _next_ state we end up as a result of processing that event.

I played with this back and forth a bit, initially states _did_ have an awareness of each other, so the LoadingState would have its own internal map stating that when a SINGLE_RESULT event fired it would lead to the SingleResultState. I moved away from this because I felt like it was coupling the states together in a way that made the individual states need too much information about other states.

In this way, states just fire off events that need to be handled elsewhere, and detail what those events could be. It's up to the manager to handle mapping those events to the appropriate states.

tldr: The mapping should be in the manager so we can modify it on the fly. The definitions should be in the state class.

Cellivar:
>  I think you need to be very clear in definition for this.

Looking at this with fresh coffee has reminded me why I was interested in the state machine in the first place: modifying the event mapping on the fly. I'll be adding the appropriate logic for that now.

src/regdesk/states/multiple_result_state.js
Cellivar: I agree, this was expedient but rubbed me the wrong way. I'm expecting to hoist most of this state information out into the commMgr (which doesn't exist yet) and have very limited info, if anything, dropped into the event object.

Cellivar:
Excuse me while I just _completely add the UI to this_. Sorry for the size, let me know if you need me to break it up differently.

To break down the changes: 

* Drops the `details` field from all CustomEvent objects for transitions. All state is managed externally, mostly through commMgr. I'm not 100% on what commMgr is actually handling or how it should handle it but consider that entire class one big mock to get the rest of the classes closer to where they need to be.
* Add the skeleton commMgr. This is unlikely to be its final form, so please feel free to drop in comments on it with the caveat that it may change drastically in the next PR.
* The state map in the state manager now uses methods to determine what the next state should be. This lets us make the next state conditional on the fly based on additional details.
* Adds an unpaid state between selecting a single search result and the print page. This gives us a clear stopping point if the station isn't set up for payments, and an API surface to mark them as having been paid.
* Adds the page UI for the various pages and wires them up. You can pull this and run it locally to get an idea of the workflow, including minor pauses for loading screens.

And some minor ones:

* Added some workflow hints to various pages, I'll do a dedicated pass to ensure that hint instructions are present at every step of the way.
* Used HTML templates for the list of multiple reg results, internet seems to be wishy-washy on whether this is a good idea.
* Save for the template cloning, all querying for state-managed elements are done in the state constructors once. Classes are re-used
* Cleaned up errant asyncs floating around.
* A to-do item is to move the eventMap details out of individual state values. At this point there is unlikely to be any reason that the individual states need to be aware of their own event map so storing it there is just vestigial, it can move out. Work for another PR, this one is massive already.

public/regdesk.html
```diff
@@ -70,30 +76,122 @@
...
           </div>
         </div>
         <div id="singleResultState" class="row d-none">
           <div id="searchResults" class="row">
-            <!-- Demonstration canvas, actual layout will be different. -->
-            <canvas id="canvas" width="589" height="193" style="background-color: white; width: 589px; height: 193px; padding: 0;"></canvas>
+            <form class="row">
+              <div class="col-md-4">
+                <label for="regPreferredName" class="form-label">Preferred Name</label>
+                <input id="regPreferredName" type="text" class="form-control form-control-lg" disabled />
```
kobaj, line 100: I'm not a big fan of these being inputs. Even if they are disabled. It implies that the data is mutable in some fashion and adds visual noise (outlines) to the interface. This could be simplified down to some flexboxes (imho) or a table.

public/regdesk.html
```diff
@@ -70,30 +76,122 @@
...
+                <div class="col-md-4">
+                  <label class="form-label">Date of Birth</label>
+                  <div class="input-group">
+                    <input type="text" name="dob" class="form-control form-control-lg" disabled />
+                    <span class="input-group-text">Age-<span name="age"></span></span>
+                  </div>
+                </div>
+                <div class="col-md-4 offset-md-4 d-grid">
+                  <label class="form-label"></label>
+                  <button type="submit" class="btn btn-lg btn-success">Select this registration</button>
```
kobaj: Going with the principle of "giant button good" we could make this entire row "clickable". 

If this extension used on a touchscreen, Keep the button, its a good visual indicator. 

If this extension is used on a regular computer (ie, has mouse-hover), then I wouldn't even bother with the button. Just make the row glow green on hover.

src/regdesk/communication_manager.js
```diff
@@ -0,0 +1,142 @@
+
+/**
+ * Manager for handling communication to the extension and APIs.
+ */
+export class CommunicationManager {
```
kobaj, line 5: I'm super guilty of this in personal projects, so I don't actually think this is a problem here. But technically "manager" is a nothing word just like "util" or "helper". That is, its not constrained enough in definition and it opens up the class to be ever expanding. 

https://stackoverflow.com/questions/1866794/naming-classes-how-to-avoid-calling-everything-a-whatevermanager

src/regdesk/communication_manager.js
```diff
@@ -0,0 +1,142 @@
...
+   * Get the string that was last searched for
+   */
+  get searchString() {
+    return this.#searchString;
+  }
+
+  /**
+   * Iniitalizes a new instance of the CommunicationManager class.
+   */
+  constructor() {
```
kobaj, line 9: I do think this class is doing /way/ too much though. Like you said, it kind of just ended up the stomping ground for state-state communication. Which should be far more regulated. 

But also like you said you'll be refactoring it, and so I'll not leave any other comments in this file. :)

src/regdesk/states/loading_state.js
```diff
@@ -52,6 +58,44 @@ export class LoadingSearchResultState extends RegState {
...
+
+    this.dispatchTransition(event);
+  }
+
+  /**
+   * Sanitize a string for HTML display, such as inside a <span>.
+   * @param {string} str - The string to sanitize
+   * @return {string} The sanitized string
+   */
+  #santizeStringForHtml(str) {
```
kobaj: As they say never roll your own encryption, never roll your own sanitization either.

https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html

src/regdesk/states/loading_state.js
```diff
@@ -52,6 +58,44 @@ export class LoadingSearchResultState extends RegState {
...
+    this.#acceptResults = false;
+  }
+
+  /**
+   * Handle the search results once they're available.
+   * @param {*} results - The results of the search.
+   */
+  handleResultsReady(results) {
+    if (!this.#acceptResults) {
+      return;
```
kobaj: I love your early returns and denested structure! Thank you for practicing this clean code strategy :D.

src/regdesk/states/loading_state.js
```diff
@@ -15,6 +15,8 @@ export class LoadingSearchResultState extends RegState {
     };
```
kobaj: Github doesn't let me leave a comment on the above lines for some reason...

This file should match the class name. `loading_search_result_state.js`

src/regdesk/states/multiple_result_state.js
```diff
@@ -22,18 +22,8 @@ export class MultipleResultState extends RegState {
...
-        MultipleResultState.events.SELECTED_SINGLE_RESULT,
-        { badgeLine1: 'ASDF', fromMultiple: true });
-    });
-    this.button2 = this.screenRow.querySelector('#multipleResultExampleTwo');
-    this.button2.addEventListener('click', () => {
-      this.dispatchTransition(
-        MultipleResultState.events.SELECTED_SINGLE_RESULT,
-        { badgeLine1: 'FDSA', fromMultiple: true });
-    });
+    this.template = this.screenRow.querySelector('#multipleResultTemplate');
```
kobaj, line 34: I personally like your use of templates, I think they could go even further with something like posthtml-modules to make it really re-usable.

src/regdesk/states/multiple_result_state.js
```diff
@@ -52,4 +58,29 @@ export class MultipleResultState extends RegState {
...
+
+  /**
+   * Handle the registration selection
+   * @param {Event} e - The submit event object
+   */
+  handleRegSelection(e) {
+    e.preventDefault();
+
+    const regId = e.target.querySelector('input[name=resultId').value;
+    this.commManager.setSelectedSearchResult(regId);
```
kobaj: I don't like this at all. It feels like you're wanting commManager to be [Redux](https://redux.js.org/) (centralized state), which is just...gross. Why get rid of passing details between states on the dispatch transition? I liked the idea, just needed to ensure what was being passed was actual data (searchResults) and not metadata (previousState).

src/regdesk/states/new_search_state.js
```diff
@@ -30,11 +30,10 @@ export class NewSearchState extends RegState {
...
     this.searchForm.addEventListener('submit', (e) => {
       e.preventDefault();
-      // Just to demonstrate the setup, this would talk to the commMgr
-      // and actually run the search.
-      // TODO: Should the event objects be strongly typed? Correlated somehow?
-      const searchText = this.searchInput.value;
-      this.dispatchTransition(NewSearchState.events.START_SEARCH, { searchText: searchText });
+
+      // Only start the search process, the loading screen waits for it.
+      this.commManager.startSearchForRegByName(this.searchInput.value);
```
kobaj: For a detailed explanation for why I don't like this and I think its is gross, its (primarily) because its a side-effect. 

https://softwareengineering.stackexchange.com/questions/15269/why-are-side-effects-considered-evil-in-functional-programming

src/regdesk/states/reg_machine_manager.js
```diff
@@ -116,4 +129,28 @@ export class RegMachineManager {
     this.#currentState = nextStateName;
     return this.#currentState;
   }
+
+  /**
+   * Determine if a cancel should go back to a multiple result page or the new search page.
+   * @return {string} - The next state to switch to.
+   */
+  #maybeCancelSingleResultToMultipleResult() {
+    if (this.#regMachineArgs.commManager.hasMultipleResults) {
```
kobaj: Oh man, I thought we talked about law of demeter?

The way I would structure this is to have a stack, each time there is a state transition just push that state onto the stack. Each time someone clicks cancel, pop the state. You could trim/clear the stack when someone clicks new search. The beauty is that this acts very much like a back button (which we should support imho, but not going to push it), and it doesn't rely on commManager. 

It gets even more beautiful if you include the details passed between states on each stack entry too. And instead of "actually" trimming/clearing the stack, you just put a marker where the "user head" is. These two minor adjustments means we can treat this stack like an audit log "these were all the searches, prints, etc that happened on this station."

src/regdesk/states/reg_machine_manager.js
```diff
@@ -116,4 +129,28 @@ export class RegMachineManager {
...
+    } else {
+      return NewSearchState.name;
+    }
+  }
+
+  /**
+   * Determine if a single result is unpaid and should get an unpaid screen.
+   * @return {string} - The next state to switch to.
+   */
+  #maybeSingleResultToUnpaid() {
```
kobaj, line 148: Hmmmmmmmmm, I don't think this method belongs here. This manager should only care about metadata (previousState) and not actual data (searchResults). 

The problem is, what I would recommend (some extra `Transition` classes that hold things like this) would be a lot of extra effort. So I think having this method here is fine for now, just keep an eye on it and if the codebase gets complex this would be an area to refactor.

src/regdesk/states/multiple_result_state.js
kobaj: Half of me is frustrated that it requires modifying two different files/places when changing state code. But I get what you're saying and I do agree with you. So for now this is probably fine, I'll keep an eye on it. :)

src/regdesk/states/unpaid_result_state.js
```diff
@@ -0,0 +1,121 @@
...
+    const reg = this.commManager.selectedSearchResult;
+
+    this.regPreferredName.value = reg.preferredName;
+    this.regLegalName.value = reg.legalName;
+    this.regBirthdate.value = reg.birthdate;
+    const age = this.#getAge(reg.birthdate);
+    this.regAge.textContent = age;
+    this.regPayment.value = reg.amountDue;
+
+    // TODO: Move this to some Registrant class?
```
kobaj, line 69: Could make the argument that the payment screen shouldn't have anything to do with the badge screen. IE, just remove this whole portion.

src/regdesk/states/unpaid_result_state.js
```diff
@@ -0,0 +1,121 @@
...
+    this.dispatchTransition(this.constructor.events.PAID);
+  }
+
+  /**
+   * Get an age in years between the birthdate and today.
+   * @param {string} birthdate - The raw birthdate string, in YYYY-mm-dd format.
+   * @param {Date} today - The 'today' value to use. Defaults to new Date().
+   * @return {number} - Age in years between the birthdate and today.
+   */
+  #getAge(birthdate, today = new Date()) {
```
kobaj, line 122: Copy past code :P

Cellivar: Yep, as the todo mentions I expect to pull this out into a different class that does that calculation in one place. The 'jumble of reg details' deserves its own class but I haven't seen precisely what the API's going to be returning.

src/regdesk/states/loading_state.js
Cellivar: This is lifted off of that page, see [rule #1](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1-html-encode-before-inserting-untrusted-data-into-html-element-content) for HTML element contexts. Apparently I mis-read it as I thought single and double quotes were kosher for this, I'll add those.

src/regdesk/communication_manager.js
Cellivar: As someone who spent the time to rename 200+ "helper" classes a number of years back I feel you. When I don't have a better descriptor for "thing that manages the state and operations of <item>" I tend to fall back on the "manager" term which is definitely underspecified. It doesn't help that "Communication" is itself underspecified.

I'll keep this in mind as I think about how to break up the commMgr class.

src/regdesk/states/unpaid_result_state.js
Cellivar: I don't disagree, but I do want to keep the badge preview on the payment screen as it's a quick and useful additional check for "This is actually you, right?" that the cashiers need to do. I don't remember if it was FC but I remember a story of two people with the same name and same birthday getting mixed up at registration. Probably the corner case of corner cases, but slapping the badge canvas up there makes it ezpz to disambiguate.

~That and I'm really proud of that code and want to throw it everywhere :V~

public/regdesk.html
Cellivar: In this situation I think the outlines are important. It gives a very clear delineation between what is the data being displayed and what data is just there as guidance or labelling, shows clear separation of fields, and provides a consistent 'anchor' for your eyes to dart to when glancing through details. 

While it doesn't look as graceful or clean as a more modern style interface I think it's easier for brains to parse after seeing it a handful of times, if that makes sense? 

The mutable element is a good call out, though I'd counter that the only mutable field in the entire system is going to be that first search field that is very obviously a search field. The lack of any 'save' or 'update' or other indicators that the forms are, well, real forms is sufficient (imo) for this purpose.

If there's more time to circle back on UI design we could see if there's a middle ground to provide that clean separation of content vs guidance that doesn't involve using forms, I'm adding a sticky note for that.

public/regdesk.html
Cellivar: I've generally been sticking to OOTB Bootstrap stuff for the UI to speed things along. It's easy enough to add a click handler to the entire list element though. Adding a CSS hover indicator probably wouldn't be too bad either.

The chromebooks we have to work with are not touchscreens and folks will have a wired mouse to use.

src/regdesk/states/multiple_result_state.js
Cellivar: It's definitely walking the line of the SRP/SoC guidance, it's fair feedback.

src/regdesk/states/multiple_result_state.js
Cellivar: A decade ago I used and quite enjoyed [using jsRender](https://www.jsviews.com/#jsr-quickstart) when I was first learning frontend dev. The clunkyness of these HTML templates makes me dislike using them more broadly, they feel like an incomplete solution without a really great way to identify the parts of the template that are meant to be substituted. This is why I started with the javascript template string HTML thing, it just felt.. nicer? Alas.

src/regdesk/states/loading_state.js
Cellivar: Oops, missed that when I renamed the class. Fixed.

src/regdesk/states/multiple_result_state.js
Cellivar: I waffled a bit on this, and pushing the selection state to a separate class generally felt like a cleaner implementation as it let the enter/exit methods be more pure. I was also itchy about mapping event details in a stronger fashion that JS doesn't.. have easy to use mechanisms for. 

Let me play with it a bit.

src/regdesk/states/new_search_state.js
Cellivar: Yeah I'm definitely familiar with pure methods and didn't enjoy writing this code. In C# I'd have a lot more message passing operations in here with immutable records being handed about and it just feels like it's an uphill battle in JS to do that cleanly.

src/regdesk/states/reg_machine_manager.js
Cellivar: >  back button (which we should support imho, but not going to push it)

I'm going to veto this purely because the interface is designed to be used in fullscreen and the mice we're using don't have forward/back buttons. Could be a future improvement though, but definitely not for this version.

Rest of your comment is good feedback, going to take a look at a state chain. I dismissed it pretty early as I wanted entry into a state to be mostly idempotent but after some evolution I don't think that matters at this point.

src/regdesk/states/loading_state.js
Cellivar: Ahha, apparently single and double quotes are okay for bodies of elements like spans but are not kosher for _attributes_ of HTML elements. That's why I didn't have them in there in the first place.

Either way fixed now.

src/regdesk/states/reg_machine_manager.js
Cellivar: So we chatted a bit about the stack design and how we could implement it, and I think the call to be made here is that the system won't benefit from the added complexity of that design.

Right now we only have a handful of cancel 'targets', that is, states where you can end up after a cancel action happens. Those states are the 'multiple select state' and the 'new search' state. All other states are transient, in that you should not see them if you hit cancel.

This means the actual contents of a 'cancel stack' are going to be very limited, and we will need to describe that relationship as a _separate_ value within each state class. This functionally would mean two parallel state transition maps: one for forward workflows, one for back workflows. This also makes 'forward' operations much more complicated, as there are specific rules about whether a state can be re-entered after exiting it. 

I think this adds _power_, however that power would not be usefully employed for the con next week. If that makes sense? A todo for version 2.

Cellivar: > Oh man, I thought we talked about law of demeter?

For this one specifically I don't think it applies. The regMachineArgs is a handy dandy transparent grouping of related arguments, not a class with behaviors unto itself, while the hasMultipleResults is an explicit item that commManager exposes.

kobaj: I can agree that a full stack is probably overpowered. But I think my argument that we shouldn't have a centralized side-effect storage class that determines state transitions still stands. But maybe you're responding to that in my other comment, so I'll sit on my hands for a bit. :)

src/regdesk/states/multiple_result_state.js
Cellivar: Alright, hauled the state out of the commMgr class back down to being passed directly between states.

What I _don't_ like about this is that it ends up tying the states together with the objects they need to pass to each other without any real great way of making that information strongly typed. I'd have to add.. a lot? of infrastructure to make it more safely typed in a way that would satisfy my strongly typed .NET programmer soul.

One of the benefits of having a 'mailbox state' object to the side of a state machine is that there is one place that everything goes to get and put its state. States don't hand details between each other, instead they put it in a well scoped location and the next state knows where to go looking for it. This lets you carefully curate the rules for what states are allowed to put there and what they're allowed to pull from there.

Obviously as soon as you add multithreaded code this all goes to hell but we aren't doing that here, hence why it was the first thing I picked up.

In moving to the state transfer between the states we've removed the mailbox (mostly, as the API stuff will still need to be there) but now we have students passing notes between each other in the class and the teacher isn't watching very closely. I'm not sure if it's a net improvement.

src/regdesk/states/reg_machine_manager.js
Cellivar: I think I'm going to call it as the PR currently stands at 9333c2b,

I've hauled the centralized state out and spread it among the various state elements, so this reduces the depth of information retrieval for the state transition determination method.

Right now the RegMachineManager is in charge of transitions between the states that it manages. I think it's fair game to say that the manager can introspect the various states under its management to determine how it should be redirecting events between the states. I don't think our complexity needs are going to grow to the point that we need another layer of abstraction on top of this, and I can already see how to fairly readily add in support for the SingleResultAlreadyPrinted state that we discussed adding today.

So I think I'm going to leave this specific bit of feedback for post-con cleanup, especially once future plans are solidified and we decide how much of this we're going to re-use.

src/regdesk/states/multiple_result_state.js
kobaj: I agree its a little funny feeling that the objects/data passed between each state is not strongly typed and yet is strongly correlated. But I would argue the non-mailbox approach is far more pure and idempotent. Just imagine trying to unit test, each arrange portion of the test methods would have 50 lines of setup for 1 line of act (followed by 50 more lines of assertions).

In this case, I would happily take some uncertainty in types compared to the simplicity of reasoning through state transitions.

public/regdesk.html
```diff
@@ -55,9 +55,15 @@
...
     <div class="row row-no-margin container-no-gutter h-100">
       <div class="col-2 d-grid gap-2 h-100 container-no-gutter-left">
-        <button id="cancelBtn" class="btn btn-danger btn-fullheight btn-huge">CANCEL</button>
+        <button id="cancelBtn" class="btn btn-danger btn-fullheight btn-huge invisible">CANCEL</button>
       </div>
-      <div class="col-8 pt-3" id="centerField">
+      <div class="col-8 pt-3 h-100 overflow-auto" id="centerField">
+        <div id="hideAfterPageLoad" class="row">
+          <div class="alert alert-warning" role="alert">
+            <h2>Search bar not loading? Try refreshing the page.</h2>
```
kobaj, line 63: Fullscreen app, won't have a refresh button :P

src/regdesk/communication_manager.js
kobaj: The class looks perfect now with a single responsibility! Its a great DAO for the API layer. :D

src/regdesk/states/multiple_result_state.js
```diff
@@ -66,10 +82,12 @@ export class MultipleResultState extends RegState {
   handleRegSelection(e) {
     e.preventDefault();
 
-    const regId = e.target.querySelector('input[name=resultId').value;
-    this.commManager.setSelectedSearchResult(regId);
-
-    this.dispatchTransition(MultipleResultState.events.SELECTED_SINGLE_RESULT);
+    // The <li> item is the enclosing element for the entire reg, select that and
+    // then query within it in case some direct element was clicked on.
+    const regId = e.target.closest('li').querySelector('input[name=resultId').value;
```
kobaj, line 87: I didn't know about closest, thats really cool!

public/regdesk.html
kobaj: Totally fair, I may take a crack at it sometime too. Don't want you to take all the HTML fun!

src/regdesk/states/multiple_result_state.js
kobaj: Interesting, I'll have to read into jsRender some I've not heard of it!

kobaj:
> I think it's fair game to say that the manager can introspect the various states under its management to determine how it should be redirecting events between the states.

That's totally fair, yeah, I'm happy with that.

---

Unrelated, how come sometimes I can reply to a comment and other times I can only "Resolve conversation"? It seems like github arbitrarily determines when I can and cannot do what I want to do (that is, just leave a comment)!! For the record this is mostly me just venting at their frustrating interface, feel free to treat that question as rhetorical.

kobaj (approved):
I think this looks really good. I hope you aren't frustrated by doing a bunch of work to make a mailbox-like solution only to reverse that work. I actually hope you're seeing how your initial intuition was right. That is to say, the original implementation of passing objects/data between states is far more testable, threadable, side-effect free, pure, and imho correct and at only minor cost.

public/regdesk.html
Cellivar: All Chromebooks come with a dedicated refresh button to replace F5 :D

Cellivar:
> Unrelated, how come sometimes I can reply to a comment and other times I can only "Resolve conversation"?

I presume you mean when you see this?

image

GitHub's UI is awful about showing that this is the _second_ instance of this conversation. It's a reply you made in your _new_ review, which is then _grouped_ into the new review for context. But since it's part of a comment chain further up you can't reply directly to it, you must go find the original context it was part of. Click on the _time_ of the comment to jump to the discussion it's actually part of.

Yes, this has been a point of contention at my work.

Cellivar:
> I hope you aren't frustrated by doing a bunch of work to make a mailbox-like solution only to reverse that work. I actually hope you're seeing how your initial intuition was right. That is to say, the original implementation of passing objects/data between states is far more testable, threadable, side-effect free, pure, and imho correct and at only minor cost.

My frustration is definitely towards JS than any work going on here. I repeatedly walk into what I think is a relatively simple pattern to build only to discover JS just.. doesn't have equivalencies to C#-isms.

One of the bigger relevant ones here is [pattern matching with switch statements](https://docs.microsoft.com/en-us/dotnet/csharp/fundamentals/functional/pattern-matching#multiple-inputs), which would vastly simplify the state tree into just a single switch statement. Or for that matter reflection on the type system, with which we _could_ completely encapsulate the transition information into classes and on app startup just construct the entire state machine in one go. No need to modify two places: the first place goes looking for the second place automatically.

My usual repertoire of shenanigans isn't translating, so I have to dredge up the rusty principles. I write so much process docs and glue code at work that this is turning out to be good exercise.

Just wish we had started 6 months ago..

PR #17, opened 2022-01-02
Feature/kobaj integration test

kobaj:
Mostly made this so I could slap in some "console.log" whenever I forget what a response object looks like. But hey, its actually pretty handy if we can get it to run. I have no idea how to set env variables for github though, hence the need for "allowFail". To run it locally just add a `.env` file with the following contents.

```
EMAIL="kobaj.g@gmail.com"
PASSWORD="hunter2"
```

### DIFF
--- a/test/regfox/regfox_api.integration.js
+++ b/test/regfox/regfox_api.integration.js
@@ -48,33 +48,36 @@ describe('regfox_api (integration testing)', () => {
   });
 
   describe('workflows', () => {
+    const COMPLETED_STATUS = 3; // https://help.regfox.com/en/articles/2343628-registration-statuses-explained
+    const TEST_NAME = 'First Last';
+    const TEST_CUSTOMER_ID = '1662788';
     let bearerToken = undefined;
     before.allowFail(async () => {
       bearerToken = (await loginForTest()).token.token;
     });
 
-    it.allowFail('searches and does not find anything', async () => {
+    it.allowFail('searches for a registrant and does not find anything', async () => {
       assert.exists(bearerToken);
 
       const results = await searchRegistrations('GUARANTEED_TO_BE_UNIQUE_UNLESS_SOMEONE_HAS_A_REALLY_WEIRD_NAME', bearerToken);
       expect(results.registrants).to.be.empty;
     });
 
-    it.allowFail('searches for a user', async () => {
+    it.allowFail('searches for a registrant', async () => {
       assert.exists(bearerToken);
 
-      const results = await searchRegistrations('First Last', bearerToken);
+      const results = await searchRegistrations(TEST_NAME, bearerToken);
       expect(results.registrants).to.have.lengthOf.at.least(1);
-      const registrant = results.registrants.find((reg) => reg.status === 3 && reg.customerId === '1662788');
+      const registrant = results.registrants.find((reg) => reg.status === COMPLETED_STATUS && reg.customerId === TEST_CUSTOMER_ID);
       expect(registrant).to.exist;
       expect(registrant.id).to.be.equal('26564608');
     });
 
-    it.allowFail('searches for a user and then gets their info', async () => {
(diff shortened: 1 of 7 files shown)


test/regfox/regfox_api.integration.js
```diff
@@ -0,0 +1,83 @@
...
+  describe('workflows', () => {
+    let bearerToken = undefined;
+    before.allowFail(async () => {
+      bearerToken = (await loginForTest()).token.token;
+    });
+
+    it.allowFail('searches and does not find anything', async () => {
+      assert.exists(bearerToken);
+
+      const results = await searchRegistrations('GUARANTEED_TO_BE_UNIQUE_UNLESS_SOMEONE_HAS_A_REALLY_WEIRD_NAME', bearerToken);
```
Cellivar, line 63: 👌🏻

test/regfox/regfox_api.integration.js
```diff
@@ -0,0 +1,83 @@
...
+      expect(results.registrants).to.be.empty;
+    });
+
+    it.allowFail('searches for a user', async () => {
+      assert.exists(bearerToken);
+
+      const results = await searchRegistrations('First Last', bearerToken);
+      expect(results.registrants).to.have.lengthOf.at.least(1);
+      const registrant = results.registrants.find((reg) => reg.status === 3 && reg.customerId === '1662788');
```
Cellivar: Worth pulling these magical values out into a const of some sort listing them off as "expected to exist" values?

Cellivar (dismissed):
Code itself LGTM, couple of nitpicks.

Cellivar:
> I have no idea how to set env variables for github though, hence the need for "allowFail".

So the way to do this is via something you can't see because you don't have admin over the repo. There's a [repository secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) system that lets you add secret values to the system and reference them within a workflow. You'd then present those secrets as ENV vars using [workflow syntax](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow).

Obviously these secrets can be read out by any workflow, so I'd prefer to create a dedicated user for this instead of re-using your personal creds.

test/regfox/regfox_api.integration.js
kobaj: Yeah, thats a good call. Let me know if this is what you had in mind or something more.

kobaj:
> So the way to do this is via something you can't see because you don't have admin over the repo. There's a repository secrets system that lets you add secret values to the system and reference them within a workflow. You'd then present those secrets as ENV vars using workflow syntax.

Cool, I figured it was something like that!

> Obviously these secrets can be read out by any workflow, so I'd prefer to create a dedicated user for this instead of re-using your personal creds.

Agreed. We can ask Ray at some point in the future.

Cellivar (dismissed):

test/regfox/regfox_api.integration.js
Cellivar: LGTM. I'm a big fane of the [Arrange - Act - Assert](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) test model, where you have a nice clear separation of "here's me putting my ducks in a row, here's where I do an actual test, and here's what I expected to happen". Part of the 'arrange' step is stating what your test data is going to be.

I've also written pre-test tests which do things like validate some precondition assumptions about test data that needs to exist that can't be generated on the fly, but I think that's overkill for what we're doing.

kobaj: Totally fair, I should stop being lazy and organize my test cases into AAA. :) 

---

I tried to follow [Google's practice](https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html) where `assert` is used for pre-test, while `expect` is used for actual test cases. Not that mocha differentiates between them, but whatever.

Cellivar (approved):

PR #18, opened 2022-01-03
Add markRegistrationComplete regfox API.

kobaj:
Followed Ray's instruction as much as possible, but as the TODOs explain I don't have authorization to change things so not sure if this actually works!

### DIFF
--- a/src/regfox/regfox_api.js
+++ b/src/regfox/regfox_api.js
@@ -2,7 +2,9 @@ import { get } from 'lodash-es';
 import { buildSearchRegistrationsBody } from './queries/regfox_graphql_search_registrations_query.js';
 import { buildAuthLoginMutationBody } from './queries/regfox_graphql_auth_login_mutation_query.js';
 import { parseRegfoxGetRegistrationResponse } from './responses/regfox_api_get_registration_info_parser.js';
+import { buildMarkRegistrationCompleteBody } from './payloads/regfox_api_mark_registration_complete_body.js';
 
+const REGFOX_MARK_REGISTRATION_COMPLETE_URL = 'https://api.webconnex.com/v1/cp/forms/${formId}/registrations/${registrationId}';
 const REGFOX_GET_REGISTRATION_URL = 'https://api.webconnex.com/v1/cp/report/regfox.com/registrant/${id}';
 const REGFOX_EXCHANGE_TOKEN_URL = 'https://api.webconnex.com/auth/exchange-token';
 const REGFOX_GRAPHQL_URL = 'https://api.webconnex.com/apollo/graphql';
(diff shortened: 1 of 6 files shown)


test/regfox/regfox_api.integration.js
```diff
@@ -87,5 +87,14 @@ describe('regfox_api (integration testing)', () => {
       expect(registrantInfo.name).to.be.equal(TEST_NAME);
       expect(`${registrantInfo.customerId}`).to.be.equal(TEST_CUSTOMER_ID);
     });
+
+    it.allowFail('marks a registrant as complete', async () => {
+      assert.exists(bearerToken);
+      const registrationInfo = await getRegistrationInfo('26564608', bearerToken);
```
Cellivar: Same as the other PR, can this magic number be a const somewhere near the top?

src/regfox/regfox_api.js
```diff
@@ -123,4 +125,41 @@ const getRegistrationInfoUrl = (id) => {
...
+ *
+ * All of these parameters can be pulled straight from the registrationInfo.
+ *
+ * @param {int} formId is the form (webpage) the user signed up with, could be hardcoded to 372652
+ * @param {int} registrationId of the registrant (**not** the Id), I think sometimes this is called the OrderId
+ * @param {int} transactionId of the registrant
+ * @param {string} id of registration (**not** the RegistrationId)
+ * @param {string} bearerToken the bearer token of the logged in user
+ */
+const markRegistrationComplete = async (formId, registrationId, transactionId, id, bearerToken) => {
```
Cellivar, line 140: This feels like it's teetering towards a destructured object input instead of a line of IDs. Not sure if 5 is the line to draw, might be okay.

src/regfox/regfox_api.js
```diff
@@ -123,4 +125,41 @@ const getRegistrationInfoUrl = (id) => {
...
+      if (get(response, 'code') === 1000) {
+        throw new Error(`Illegal bearerToken passed in to markRegistrationComplete: ${JSON.stringify(response)}`);
+      }
+
+      return response;
+    });
+};
+
+// Only exported for testing, DONT CALL THIS METHOD!
+const getMarkRegistrationCompleteUrl = (formId, registrationId) => {
```
Cellivar: Would it be prudent to put the word `test` explicitly in this export then?

src/regfox/regfox_api.js
```diff
@@ -123,4 +125,41 @@ const getRegistrationInfoUrl = (id) => {
...
+/**
+ * Marks a registrant (technically a transaction of an order of a registered user) as completed (which means fully paid).
+ * See https://gist.github.com/raytingtw/8d0a083dc130c40818d71606e41b884e.
+ * Returns a promise Error if the network is down.
+ *
+ * All of these parameters can be pulled straight from the registrationInfo.
+ *
+ * @param {int} formId is the form (webpage) the user signed up with, could be hardcoded to 372652
+ * @param {int} registrationId of the registrant (**not** the Id), I think sometimes this is called the OrderId
+ * @param {int} transactionId of the registrant
```
Cellivar: Silly fact: No int type in JS means this isn't valid jsdoc apparently! 

```suggestion
 * @param {number} formId is the form (webpage) the user signed up with, could be hardcoded to 372652
 * @param {number} registrationId of the registrant (**not** the Id), I think sometimes this is called the OrderId
 * @param {number} transactionId of the registrant
```

Cellivar (dismissed):
Code LGTM. We can do some live testing once we get the APIs wired up and have some test accounts to mess with.

test/regfox/regfox_api.integration.js
kobaj: Sure!

src/regfox/regfox_api.js
kobaj: Interesting, I'm very much used to [checkstyle default maximum](https://checkstyle.sourceforge.io/apidocs/src-html/com/puppycrawl/tools/checkstyle/checks/sizes/ParameterNumberCheck.html#line.125) of 7 arguments before switching to an object.

I assumed eslint was the same, but it appears they actually set [the default max](https://eslint.org/docs/rules/max-params) to 3! The reason we aren't getting a notification on that is because we are using Google Style which has the [check turned off](https://github.com/google/eslint-config-google/blob/master/index.js#L244)!

Well I'm happy to set the max params to 5, seems like a reasonable middle ground between all three of those options.

src/regfox/regfox_api.js
kobaj: Good call, I'll make that change!

src/regfox/regfox_api.js
kobaj: Huh! Thanks for the heads up :D.

src/regfox/regfox_api.js
kobaj: Just kidding, set it to 6 so that your constructor here doesn't trigger a failure. 

https://github.com/anthroarts/RegExtension/blob/e7b8757a6b700b9a94070aa8e69917821521b4bf/src/regdesk/label_builder.js#L37

Cellivar (approved):

PR #20, opened 2022-01-03
Add manual label print button

Cellivar:
Adds an interface for manually entering arbitrary badge details.

I'm sure this won't be used to generate hilarious badges.

image

Interface is similar to the KIT badge, click the Adult printer dropdown and it's there. 

image

### DIFF
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -42,6 +42,8 @@
                 <span id="adultPrintIcon">❌</span> Adult <span id="adultPrintName">(No Printer)</span>
               </a>
               <ul class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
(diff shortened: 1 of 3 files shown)


public/regdesk.html
```diff
@@ -42,6 +42,8 @@
                 <span id="adultPrintIcon">❌</span> Adult <span id="adultPrintName">(No Printer)</span>
               </a>
               <ul class="dropdown-menu" aria-labelledby="navbarDropdownMenuLink">
+                <li id="adultPrintManual"><a class="dropdown-item" href="#" data-bs-toggle="modal" data-bs-target="#manualPrintModal">Manual Label</a></li>
```
kobaj: Should this be an action to match all the other items in the list? `Print Manual Label`

Cellivar: Good point, slightly modified to "Make" as you don't directly print something, you open a form to make the label and then click print.

kobaj (approved):

PR #21, opened 2022-01-04
Add printer config modal

Cellivar:
Adds a modal dialog for configuring printer details. The printer config is limited to relevant config items that might need to be changed or validated at the con. 

This wires the previously added config buttons that weren't doing anything.

image

### DIFF
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -240,9 +240,9 @@
                 </div>
               </div>
               <div class="col-md-4">
-                <label for="printerSettingsDarkness" class="col-sm-6 col-form-label">Darkness</label>
+                <label for="printerSettingsDarkness" class="form-label">Print Darkness</label>
                 <select id="printerSettingsDarkness" class="form-select" aria-label="Darkness">
-                  <option>1</option>
+                  <option value="1">1 - Lightest</option>
                   <option>2</option>
                   <option>3</option>
                   <option>4</option>
@@ -256,16 +256,16 @@
                   <option>12</option>
                   <option>13</option>
                   <option>14</option>
-                  <option>15</option>
+                  <option value="15">15 -Darkest</option>
                 </select>
               </div>
               <div class="col-md-4">
-                <label for="printerSettingsSpeed" class="col-sm-5 col-form-label">Speed</label>
+                <label for="printerSettingsSpeed" class="form-label">Print Speed</label>
                 <select id="printerSettingsSpeed" class="form-select" aria-label="Speed">
-                  <option>1</option>
-                  <option>2</option>
-                  <option>3</option>
-                  <option>4</option>
+                  <option value="1">1.5ips</option>
+                  <option value="2">2.0ips</option>
+                  <option value="3">2.5ips</option>
+                  <option value="4">3.5ips</option>
                 </select>
               </div>
             </div>
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -197,6 +197,74 @@
       </div>
     </div>
   </div>
+  <div id="printerOptionModal" class="modal" tabindex="-1">
+      <div class="modal-dialog">
+          <div class="modal-content">
+              <form id="printerSettingsForm">
+                  <div class="modal-header">
+                    <h5 class="modal-title">Printer Settings For: <span id="printerSettingsSerial"></span></h5>
+                    <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+                  </div>
+                  <div class="modal-body">
+                    <div class="row">
+                      <div class="col-md-12 alert alert-danger">
+                        Do not change the configuration when the label dispenser is engaged. It will dispense many labels and not stop. Ensure dispenser is disengaged before saving.
+                      </div>
+                      <div class="col-md-6">
+                        <label for="printerSettingsLabelWidth" class="form-label">Label Width</label>
+                        <div class="input-group mb-3">
+                            <input id="printerSettingsLabelWidth" type="text" class="form-control" placeholder="2.25"
+                                aria-label="Width" required pattern="\d+\.*\d*">
+                            <span class="input-group-text">in</span>
+                        </div>
+                      </div>
+                      <div class="col-md-6">
+                        <label for="printerSettingsLabelHeight" class="form-label">Label Height</label>
+                        <div class="input-group mb-3">
+                            <input id="printerSettingsLabelHeight" type="text" class="form-control" placeholder="1.25"
+                                aria-label="Height" required pattern="\d+\.*\d*">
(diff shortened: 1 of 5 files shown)


public/regdesk.html
```diff
@@ -199,7 +209,77 @@ <h4>Payment required</h4>
...
+                <label for="printerSettingsLabelHeight" class="form-label">Label Height</label>
+                <div class="input-group mb-3">
+                  <input id="printerSettingsLabelHeight" type="text" class="form-control" placeholder="1.25"
+                    aria-label="Height" required pattern="\d+\.*\d*">
+                  <span class="input-group-text">in</span>
+                </div>
+              </div>
+              <div class="col-md-4">
+                <label for="printerSettingsDarkness" class="col-sm-6 col-form-label">Darkness</label>
+                <select id="printerSettingsDarkness" class="form-select" aria-label="Darkness">
```
kobaj, line 244: You might put some kind of human connotation to these options.

```
<option value=1>darkest</option>
<option>2</option>
...
<option value=15>lightest</option>
```

public/regdesk.html
```diff
@@ -199,7 +209,77 @@ <h4>Payment required</h4>
...
+        <form id="printerSettingsForm">
+          <div class="modal-header">
+            <h5 class="modal-title">Printer Settings For: <span id="printerSettingsSerial"></span></h5>
+            <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
+          </div>
+          <div class="modal-body">
+            <div class="row">
+              <div class="col-md-12 alert alert-danger">
+                Do not change the configuration when the label dispenser is engaged. It will dispense many labels and
+                not stop. Ensure dispenser is disengaged before saving.
```
kobaj, line 224: What does this mean? There isn't a button called "disengage". Is a human supposed to interpret that to mean "while printing"? Perhaps it would be simpler to say "while printing".

public/regdesk.html
```diff
@@ -199,7 +209,77 @@ <h4>Payment required</h4>
...
+                  <option>11</option>
+                  <option>12</option>
+                  <option>13</option>
+                  <option>14</option>
+                  <option>15</option>
+                </select>
+              </div>
+              <div class="col-md-4">
+                <label for="printerSettingsSpeed" class="col-sm-5 col-form-label">Speed</label>
+                <select id="printerSettingsSpeed" class="form-select" aria-label="Speed">
```
kobaj, line 264: Speed of what? Printing? Label Dispensing? Data transfer?

public/regdesk.html
```diff
@@ -199,7 +209,77 @@ <h4>Payment required</h4>
...
+                  <option>1</option>
```
kobaj: Again, human connotation would be helpful.

```
<option value=1>fastest</option>
<option>2</option>
...
<option value=4>slowest</option>
```

src/regdesk/printer_config_modal.js
```diff
@@ -0,0 +1,64 @@
...
+   * @param {CustomEvent} e - The event object
+   */
+  async #updatePrinterConfig(e) {
+    e.preventDefault();
+    this.modal.hide();
+
+    this.printer.labelWidth = (parseFloat(this.printerLabelWidth.value));
+    this.printer.labelHeight = (parseFloat(this.printerLabelHeight.value));
+    await this.printer.configDensity(this.printerDarkness.value);
+    await this.printer.configSpeed(this.printerSpeed.value);
```
kobaj, line 59: Perhaps we should parseInt?

src/regdesk/printer_config_modal.js
```diff
@@ -0,0 +1,64 @@
...
+    await this.printer.configPrintDirection();
+    await this.printer.configLabelWidth();
```
kobaj, line 61: This method doesn't take the width as an argument?

src/regdesk/printer_config_modal.js
Cellivar: Not necessary here for the density or speed, as the values are enumerated in the selects.

public/regdesk.html
Cellivar: This is the boundary where we expect someone to know what they're doing when reading this. This page here will feature in the help documentation I've started writing that will be in a physical folder with the printers.

image

src/regdesk/printer_config_modal.js
Cellivar: Nope, it updates the printer with whatever you set the width to previously, in this case on line 56. I plan to revisit the public interface on my WebZLP library at some point to clean up some of the oddities I've run into while using it.

public/regdesk.html
Cellivar: Good point, "Print Speed" it is.

Cellivar:
I've been working with these printers so much I'm blind to their settings being unparseable for anyone else, good call on that. I've updated them to be the real inches per second value and added hints to the darkness.

public/regdesk.html
```diff
@@ -240,9 +240,9 @@ <h5 class="modal-title">Printer Settings For: <span id="printerSettingsSerial"><
                 </div>
               </div>
               <div class="col-md-4">
-                <label for="printerSettingsDarkness" class="col-sm-6 col-form-label">Darkness</label>
+                <label for="printerSettingsDarkness" class="form-label">Print Darkness</label>
                 <select id="printerSettingsDarkness" class="form-select" aria-label="Darkness">
-                  <option>1</option>
+                  <option value="1">1 - Lightest</option>
                   <option>2</option>
```
kobaj, line 246: Only just realizing, should these have a default value too?

public/regdesk.html
```diff
@@ -240,9 +240,9 @@ <h5 class="modal-title">Printer Settings For: <span id="printerSettingsSerial"><
                 </div>
               </div>
               <div class="col-md-4">
-                <label for="printerSettingsDarkness" class="col-sm-6 col-form-label">Darkness</label>
+                <label for="printerSettingsDarkness" class="form-label">Print Darkness</label>
                 <select id="printerSettingsDarkness" class="form-select" aria-label="Darkness">
```
kobaj, line 244: Are these settings retrievable in some form? So that someone can read how a printer is currently configured when they pull up this screen?

kobaj (approved):
Totally forgot to include two major comments this time. My bad. They could be added in a later PR though. :)

public/regdesk.html
Cellivar: The value should always be pulled from the printer configuration and that internally sets a default, I don't think we need one here.

public/regdesk.html
Cellivar: This interface would be the way to find them, yes.

PR #23, opened 2022-01-04
Remove jquery usage from content_script.js

binaryf0x:


Cellivar (approved):
As a straight conversion from jQuery to vanilla JS this looks fine. I've got plenty of other nitpicks with this code but I don't think it's what we'll end up using long term, only for the legacy print option juuuuuust in case.

Thanks for converting it!

PR #24, opened 2022-01-04
Remove jquery from npm dependencies.

kobaj:
A follow up to https://github.com/anthroarts/RegExtension/pull/23


Cellivar (approved):
Yesssss, make that security alert go away

PR #25, opened 2022-01-04
Add toggle payments button to regfox comm dropdown

Cellivar:
This turns the top-left element into a dropdown so we can add communication relevant options. My first mockup had this button straight up on the navbar, but in considering it I think that's _too_ easy to find and use. Eschewing the payment enable toggle behind another dropdown makes it much more difficult to hit accidentally.

In the same way that the buttons in the top right are the purview of the Printer Manager, the button in the top left is the purview of the Communication Manager, and any comms-related things (like a button to re-log-into RegFox or something) should go there.

The toggle itself is very clearly off when off, and very clearly on when on.

F0Q0Z28zrg

### DIFF
--- a/src/regdesk/communication_manager.js
+++ b/src/regdesk/communication_manager.js
@@ -1,14 +1,29 @@
+import { TogglePaymentsBtn } from './toggle_payments_btn.js';
 
 /**
  * Manager for handling communication to the extension and APIs.
  */
 export class CommunicationManager {
+  #acceptingPayments;
+  #togglePaymentsBtn;
+
+  /**
+   * Indicates whether this terminal is accepting payments.
+   */
+  get acceptingPayments() {
+    return this.#acceptingPayments;
+  }
+
   /**
    * Iniitalizes a new instance of the CommunicationManager class.
+   * @param {TogglePaymentsBtn} togglePaymentsBtn - Button to toggle payment acceptance
    */
-  constructor() {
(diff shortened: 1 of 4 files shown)


src/regdesk/communication_manager.js
```diff
@@ -58,6 +73,14 @@ export class CommunicationManager {
     return results;
   }
 
+  /**
+   * Toggle the state of accepting payments.
+   * @param {CustomEvent} e - The event object
+   */
+  #togglePayments(e) {
```
kobaj, line 80: Not sure if this is triggering a re-draw or not. If someone is on the screen and toggles the button, they may need to "refresh" the page to see the updates.

kobaj (approved):
I'm kinda iffy about this being so freely/easily clickable. Not that any of the volunteers would click it. But eh, can always move it somewhere else at another time.

src/regdesk/communication_manager.js
Cellivar: It's intentionally not, as the only place we check the payment status is on state entry. The expectation is nobody should be flipping this setting on the fly, just once per day on setup.

Cellivar:
> I'm kinda iffy about this being so freely/easily clickable.

I concur, it's why it's buried behind a menu at least. It's slightly more roadbump than previously existed for printing badges without accepting payment so I think it'll end up being okay for this year at least.

PR #26, opened 2022-01-05
Feature/kobaj add note api

kobaj:
This is how we intend on marking a registration as like "checked in" right? IE, the message would be something like "Badge Printed!" maybe with some metadata about which station or printer did the operation. Do I understand it correctly?

### DIFF
--- a/src/regfox/payloads/regfox_api_add_note.js
+++ b/src/regfox/payloads/regfox_api_add_note.js
@@ -1,0 +1,18 @@
+
+/**
+ * @return {*} a body object that can add a note to a registration.
+ *
+ * @param {string} message that should be seen as a note
+ * @param {string} id of registration (**not** the RegistrationId)
+ */
+const buildAddNoteBody = (message, id) => {
+  // I have no idea what these values mean and can't find any documentation on them.
+  const noteType = 5;
+  const createdBy = 144247; // Seems we can set this to any value?
+  const visibility = 2;
+  const editing = true;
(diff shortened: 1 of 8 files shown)


Cellivar:
> This is how we intend on marking a registration as like "checked in" right?

There's actually a separate thing for that: 

image

And a (real! public!) API for doing so: https://docs.webconnex.io/api/v2/#check-in-by-id

This will be useful for marking reprints though!

src/regfox/payloads/regfox_api_add_note.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @return {*} a body object that can add a note to a registration.
+ *
+ * @param {string} message that should be seen as a note
+ * @param {string} id of registration (**not** the RegistrationId)
+ */
+const buildAddNoteBody = (message, id) => {
+  // I have no idea what these values mean and can't find any documentation on them.
+  const noteType = 5;
+  const createdBy = 144247; // Seems we can set this to any value?
```
Cellivar, line 11: Maybe the UserID of the person creating the note?

Cellivar (approved):
Thanks for adding this!

src/regfox/payloads/regfox_api_add_note.js
kobaj: Not sure how to get that at the moment. But I'll make an action item to look into it!

PR #29, opened 2022-01-05
Add RegFox page print button

Cellivar:
This adds a button directly in the RegFox UI to print a badge label for a registrant.

image

This is not tested for _multiple_ registrants in a single transaction, that'll happen tomorrow.

The underlying framework for this is passing a message from a content script over to the RegDesk, this means we expect to have both RegFox and RegDesk open in two tabs on any cashier machine.

### DIFF
--- a/src/detail_print.js
+++ b/src/detail_print.js
@@ -1,0 +1,113 @@
+console.debug('Injected detail print script');
+
+import { sendBackgroundScriptAMessage, MESSAGE_TYPE } from './coms/communication.js';
+
+/**
+ * Wait for an element to be present on the page.
+ * @param {string} selector - The CSS selector to wait for
+ * @return {Element} The element being waited for
+ */
+async function waitForElement(selector) {
+  let element = document.querySelector(selector);
+  while (!element) {
+    element = document.querySelector(selector);
+    await new Promise((r) => setTimeout(r, 200));
+  }
+  return element;
+}
+
+/**
+ * Returns a value indicating whether the payment is complete.
+ * @return {boolean} Whether payment is completed
+ */
+function isPaymentCompleted() {
+  const payStatus = document.querySelector('tr[ng-if*="report.statusString"] > td').textContent;
+  return payStatus === 'Completed';
+}
+
+/**
+ * Get the details for a reg data table entry.
+ * @param {string} fieldName - The text field name from the reg details table
+ * @return {string} The value of the table
+ */
+function getDetail(fieldName) {
+  const regDetailTable = document.querySelector('div[ng-include*="data.html"] > table');
+  const th = [...regDetailTable.querySelectorAll('th')]
+    .filter((el) => el.textContent.indexOf(fieldName) > -1);
+
+  if (th.length > 0) {
+    return th[0]?.nextElementSibling.textContent ?? '';
+  }
+  return '';
+}
+
+/**
+ * Get an age in years between the birthdate and today.
+ * @param {string} birthdate - The raw birthdate string, in YYYY-mm-dd format.
+ * @param {Date} today - The 'today' value to use. Defaults to new Date().
+ * @return {number} - Age in years between the birthdate and today.
+ */
+function getAge(birthdate, today = new Date()) {
+  // TODO: This method should move to a dedicated Registrant class.
+  const yearInMs = 3.15576e+10;
+  return Math.floor((today - new Date(birthdate).getTime()) / yearInMs);
+}
+
+/**
+ * Get the details for a badge.
+ * @return {*} The details to populate a BadgeLabelBuilder
+ */
+function getBadgeDetails() {
+  const dob = getDetail('Date of Birth');
+  const age = getAge(dob);
+  return {
+    line1: getDetail('Badge Line 1 Text'),
+    line2: getDetail('Badge Line 2 Text'),
+    level: getDetail('Membership Levels'),
+    badgeId: document.location.href.split('/').pop(),
+    isMinor: (age < 18),
+  };
+}
+
+/**
+ * Send a request to print a badge to the RegDesk page and printers.
+ */
+function printBadge() {
+  if (!isPaymentCompleted()) {
+    // Most reliable way to ask the user..
+    const bypassPayment = window.confirm('The payment status in not "Completed". Are you sure you want to print this badge?');
+    if (!bypassPayment) {
+      return;
+    }
+  }
+
+  const labelDetails = getBadgeDetails();
+  console.log(labelDetails);
+
+  sendBackgroundScriptAMessage(
+    MESSAGE_TYPE.printLabel,
+    { labelDetails: labelDetails },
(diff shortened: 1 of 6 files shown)


src/detail_print.js
```diff
@@ -0,0 +1,113 @@
...
+import { sendBackgroundScriptAMessage, MESSAGE_TYPE } from './coms/communication.js';
+
+/**
+ * Wait for an element to be present on the page.
+ * @param {string} selector - The CSS selector to wait for
+ * @return {Element} The element being waited for
+ */
+async function waitForElement(selector) {
+  let element = document.querySelector(selector);
+  while (!element) {
```
kobaj: Nitpick, I try to avoid infinite loops where possible. Should this throw an exception after like 10 attempts?

src/detail_print.js
```diff
@@ -0,0 +1,113 @@
...
+  return element;
+}
+
+/**
+ * Returns a value indicating whether the payment is complete.
+ * @return {boolean} Whether payment is completed
+ */
+function isPaymentCompleted() {
+  const payStatus = document.querySelector('tr[ng-if*="report.statusString"] > td').textContent;
+  return payStatus === 'Completed';
```
kobaj, line 27: Shouldn't really trust a text string that could change. Is the integer value not available? https://help.regfox.com/en/articles/2343628-registration-statuses-explained

src/detail_print.js
```diff
@@ -0,0 +1,113 @@
...
+ * @return {*} The details to populate a BadgeLabelBuilder
+ */
+function getBadgeDetails() {
+  const dob = getDetail('Date of Birth');
+  const age = getAge(dob);
+  return {
+    line1: getDetail('Badge Line 1 Text'),
+    line2: getDetail('Badge Line 2 Text'),
+    level: getDetail('Membership Levels'),
+    badgeId: document.location.href.split('/').pop(),
```
kobaj, line 69: Is that what we should call this id? Alright, I'll use that going forward. :)

src/event_page.js
```diff
@@ -16,7 +16,22 @@ const getAllRegfoxTabs = async () => {
   return chrome.tabs.query(queryOptions);
 };
 
+const injectPrintButtonScript = (tabId, change) => {
+  // Change event only includes the details that actaully changed. We only want
+  // to inject this script
```
kobaj: incomplete sentence?

src/detail_print.js
```diff
@@ -0,0 +1,113 @@
...
+}
+
+/**
+ * Get an age in years between the birthdate and today.
+ * @param {string} birthdate - The raw birthdate string, in YYYY-mm-dd format.
+ * @param {Date} today - The 'today' value to use. Defaults to new Date().
+ * @return {number} - Age in years between the birthdate and today.
+ */
+function getAge(birthdate, today = new Date()) {
+  // TODO: This method should move to a dedicated Registrant class.
```
kobaj, line 53: staaahhhhp

src/event_page.js
```diff
@@ -16,7 +16,22 @@ const getAllRegfoxTabs = async () => {
...
 };
 
+const injectPrintButtonScript = (tabId, change) => {
+  // Change event only includes the details that actaully changed. We only want
+  // to inject this script
+  if (!change?.url?.match(/manage\.webconnex\.com\/a\/\d+\/reports\/orders\/\d+\/details/)) {
+    return;
+  }
+
+  chrome.scripting.executeScript({
```
kobaj, line 26: Why not make this a real content script? https://developer.chrome.com/docs/extensions/mv3/content_scripts/

src/regdesk/communication_manager.js
```diff
@@ -73,6 +84,48 @@ export class CommunicationManager {
...
+  #handleExtensionMessage({ type, payload }, sender, callback) {
+    console.debug('Received extension message', type, payload, sender);
+
+    switch (type) {
+    case MESSAGE_TYPE.printLabel:
+      this.#handleExtensionPrintRequest(payload?.labelDetails).then(callback);
+      break;
+    }
+
+    return true;
```
kobaj, line 126: Do we need to trigger the callback manually when the switch/case doesn't match?

src/event_page.js
Cellivar: I tried that at first, but due to the single-page-app shenanigans that RegFox is doing I had a really hard time getting a lock on any of the events that would make the buttons show up without just.. constantly running a while loop in the background looking for elements. Because Extension scripts are locked in a different 'world' than the on-page JS we can't really listen in directly, we can only observe the effects those scripts perform.

Well, content scripts can't listen into the tab events. The tab events tell us exactly what we need to know: when we end up on a detail URL. Thus it seemed like the "most" elegant out of the options available.

src/regdesk/communication_manager.js
Cellivar: We don't necessarily know what the callback was for or what data it wants. Safer I would think to just drop it on the floor and hope that upstream has some kind of way to tell it didn't get through.

src/detail_print.js
Cellivar: I'm not fully versed on the various IDs that RegFox uses, but I know for a fact that whatever that ID is we treat it as the ID we place on badges. I suspect it's the transaction ID.

src/detail_print.js
Cellivar: I agree I don't like how fragile that is. Unfortunately the only other option appears to be reading out the class list and doing a text compare on that:

image

The failure mode here is "we accidentally warn someone to double check when the payment is actually completed" which I think is an acceptable failure, which is why I did an exact compare against the single pass-through case we'd want.

src/regdesk/communication_manager.js
kobaj: That's fair. Should we only return true in the switch case then?

src/detail_print.js
Cellivar: If I promise the next PR will include a refactor that moves all 4 (!) of these to a single spot will you let this one through? :)

src/event_page.js
Cellivar: I feex.

src/detail_print.js
Cellivar: Added a reasonable limit of 8 seconds worth of attempts. I've had weird hiccups loading the page but this seems to give it enough leeway to load enough results to get the button arranged on the page successfully.

kobaj (approved):

src/regdesk/communication_manager.js
Cellivar: The 'true' here only indicates that we want to respond asynchronously to the incoming message, it doesn't get passed to the actual sender. Either way (async or not) the sender won't receive a response.

It's possible we could leak _something_ by saying we'll respond async but don't, but the docs don't seem to indicate it's a problem? I'll keep an eye out for weirdness during testing to see if anything comes of it.

PR #30, opened 2022-01-06
Add checkIn regfox api.

kobaj:
I saw the official API here and started that implementation, 

https://docs.webconnex.io/api/v2/#check-in-by-id

But right at the end noticed it takes in an API key rather than a bearer token. After much internal debate, I decided consistency was more important at the moment. So I'm using the front-end API instead.

### DIFF
--- a/test/regfox/regfox_api.test.js
+++ b/test/regfox/regfox_api.test.js
@@ -7,13 +7,20 @@ const { expect } = chai;
 
 import { stub } from 'sinon';
 
-import { exchangeBearerTokenHandler, getRegistrantsSearchHandler, loginHandler, getRegistrationInfoHandler, markRegistrationCompleteHandler, addNoteHandler } from '../mocks/mws_handler.js';
+import {
+  exchangeBearerTokenHandler, getRegistrantsSearchHandler, loginHandler,
+  getRegistrationInfoHandler, markRegistrationCompleteHandler,
+  addNoteHandler, checkInHandler,
+} from '../mocks/mws_handler.js';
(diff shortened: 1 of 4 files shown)


Cellivar:
>  After much internal debate, I decided consistency was more important at the moment. So I'm using the front-end API instead.

SGTM honestly, the less dealing with auth requirements the better.

test/regfox/regfox_api.test.js
```diff
@@ -250,4 +257,43 @@ describe('regfox_api', () => {
...
+        code: 1000,
+        message: 'bad request',
+      });
+
+      return checkIn(ID, 1234).should.eventually.be.rejected;
+    });
+
+    it('handles a duplicate request', async () => {
+      getResponse.returns({
+        code: 1000,
```
Cellivar, line 290: Is it really the same code for both?

Cellivar (approved):
LGTM. In your explorations did you discover any method to get whether someone is checked in already easily? Does it come with the big bank of information we get from a registration?

test/regfox/regfox_api.test.js
kobaj: Yes. that surprised me at first too. You can see the commit [here](https://github.com/anthroarts/RegExtension/commit/5e9992011e855b019ebe5db60ccf0d3bb25f0ad1#diff-a4e9c7b0222ec040a3618347a2b511d482a5dc812c6f66b21f924e7caec8c4f1R174) where I discovered "Oh no, they use code 1000 to mean everything."

PR #31, opened 2022-01-06
Add RegDetails Class, wire it up along with printing

Cellivar:
Here's the class I promised would deal with the code clones of the getAge method :D

This adds a class that wraps up all the random details of a single registration in a bundle for movement around. Eventually the API hits will be coerced into this format as well, where appropriate to do so.

The various places where this information should be displayed were modified to use it. Since I was in there I also slightly tweaked the Print button so it would be possible to actually wire it up. 

RegFox print works, RegDesk print works.

### DIFF
--- a/src/detail_print.js
+++ b/src/detail_print.js
@@ -1,6 +1,7 @@
 console.debug('Injected detail print script');
 
 import { sendBackgroundScriptAMessage, MESSAGE_TYPE } from './coms/communication.js';
+import { RegistrantDetails } from './regdesk/registrant_details.js';
 
 /**
  * Wait for an element to be present on the page.
@@ -19,22 +20,13 @@ async function waitForElement(selector) {
 }
 
 /**
- * Returns a value indicating whether the payment is complete.
- * @return {boolean} Whether payment is completed
- */
-function isPaymentCompleted() {
-  const payStatus = document.querySelector('tr[ng-if*="report.statusString"] > td').textContent;
-  return payStatus === 'Completed';
-}
-
-/**
  * Get the details for a reg data table entry.
  * @param {string} fieldName - The text field name from the reg details table
  * @return {string} The value of the table
  */
 function getDetail(fieldName) {
-  const regDetailTable = document.querySelector('div[ng-include*="data.html"] > table');
-  const th = [...regDetailTable.querySelectorAll('th')]
+  // const regDetailTable = document.querySelector('div[ng-include*="data.html"] > table');
+  const th = [...document.querySelectorAll('th')]
     .filter((el) => el.textContent.indexOf(fieldName) > -1);
 
   if (th.length > 0) {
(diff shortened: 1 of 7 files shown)


src/detail_print.js
```diff
@@ -18,23 +19,14 @@ async function waitForElement(selector) {
...
-
 /**
  * Get the details for a reg data table entry.
  * @param {string} fieldName - The text field name from the reg details table
  * @return {string} The value of the table
  */
 function getDetail(fieldName) {
-  const regDetailTable = document.querySelector('div[ng-include*="data.html"] > table');
-  const th = [...regDetailTable.querySelectorAll('th')]
+  // const regDetailTable = document.querySelector('div[ng-include*="data.html"] > table');
```
kobaj: Nitpick, this will be stored in git history if we ever need it. Should just delete the whole line  rather than leaving commented out code laying around.

src/regdesk/registrant_details.js
```diff
@@ -0,0 +1,121 @@
...
+   * @param {string} o.birthdate - Birthday, in yyyy-MM-dd format.
+   * @param {string} o.amountDue - The remaining amount due.
+   * @param {string} o.badgeLine1 - The line 1 of the badge.
+   * @param {string} o.badgeLine2 - The line 2 of the badge.
+   * @param {string} o.sponsorLevel - The sponsorship level.
+   * @param {string} o.badgeId - The registrant ID to use as the badge ID.
+   * @param {string} o.conbookCount - The number of con books in the order.
+   * @param {string} o.checkinDate - The check in date string.
+   * @param {string} o.paymentStatus - The status string for the transaction.
+   */
```
kobaj, line 42: `numberOfBadgesPrintedSoFar` or similar a good idea?

src/regdesk/communication_manager.js
```diff
@@ -107,22 +107,32 @@ export class CommunicationManager {
...
    * @param {*} o.payload - The request payload.
-   * @param {*} sender - The sender of the event.
+   * @param {chrome.runtime.MessageSender} sender - The sender of the event.
    * @param {function(*)} callback - Callback message to response to the event sender.
    * @return {boolean} - True to indicate this will response async.
    */
   #handleExtensionMessage({ type, payload }, sender, callback) {
     console.debug('Received extension message', type, payload, sender);
 
+    // NOTE: Any object that is passed through the message system will STRIP all
```
kobaj, line 120: The only methods are getters, eg `hasBeenCheckedIn` why not make those variables set by the constructor? Then we don't need to have this big clumsy copy going on.

src/regdesk/states/single_result_state.js
```diff
@@ -71,17 +70,15 @@ export class SingleResultState extends RegState {
...
   }
 
   /**
-   * Get an age in years between the birthdate and today.
-   * @param {string} birthdate - The raw birthdate string, in YYYY-mm-dd format.
-   * @param {Date} today - The 'today' value to use. Defaults to new Date().
-   * @return {number} - Age in years between the birthdate and today.
+   * Print the label from the latest result.
    */
-  #getAge(birthdate, today = new Date()) {
```
kobaj, line 82: Yay! I'm super glad this method was cleaned up. Thank you for your commitment to clean code by reducing copy pastes :D.

src/regdesk/registrant_details.js
Cellivar: For our current purposes I both don't know that there's a really reliable way to pull this (the notes field is _complicated_) and I don't know that the actual quantity is useful. We're pulling in the checkinDate as our indicator that "this person has been checked in" and I think that's good enough for the moment.

src/regdesk/communication_manager.js
Cellivar: Therein lies the fun: It's recursive. The methods on the label were also removed, which meant the label that is generated wasn't able to print. Originally the getter for the label was a method called `getLabel()` as I don't really like getters that call constructors. The workaround didn't work :/

Rather than go down the rabbit hole of why this is happening I made a judgement call on limiting my deep knowledge of Chrome/JS internals and just made a copy constructor. It's absolutely the brute force solution but it's also the most reliable.

src/detail_print.js
Cellivar: Left in by mistake while testing, removed!

src/regdesk/communication_manager.js
kobaj: You and I have different definitions of reliable :P

kobaj (approved):

PR #37, opened 2022-01-30
Add html/layout for a Login Modal.

kobaj:
I figured this should be a Modal because if someone is in the middle of checking a registrant in, we don't want them to lose their work. IE, I don't think it should be part of the state diagram.

I also made it "optional" (IE "Work Offline") because if regfox goes down, we want to still be able to access the manual print mode.

Still lots of work to do to wire it in, and make triggering it better. 

image

### DIFF
--- a/public/regdesk.html
+++ b/public/regdesk.html
@@ -17,8 +17,8 @@
     <nav class="navbar navbar-expand-lg navbar-dark bg-dark">
       <div class="container-fluid container-no-gutter d-flex">
         <div class="navbar-nav nav-item dropdown">
-          <a class="nav-link dropdown-toggle active" href="#" id="connectConfig" role="button"
-            data-bs-toggle="dropdown" data-bs-auto-close="outside" aria-expanded="false">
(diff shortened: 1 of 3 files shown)


Cellivar (changes requested):
Code itself LGTM! Thanks for following the other patterns in the codebase.

Only change is that I'd like to see the offline mode button moved to the left side of the modal instead of adjacent to the login button. Grouped buttons like that are most commonly associated with "cancel | confirm" type dialogs, while the offline button is a _mode switch_ instead of a go back _navigation_. I think visually distancing it from the other mode makes it more clear that you're doing something very different than logging in.

There's some Bootstrap classes that should facilitate this.

kobaj:
Totally fair callout, going along those lines I've made the modal slightly more "mandatory". IE, you need to deliberately click Work Offline or Login. Hitting esc on the keyboard or clicking on the background will do nothing now.

image

Cellivar (approved):
Nice!

PR #38, opened 2022-02-02
Add loginManager and tests.

kobaj:
Lookit, you can refresh the page now and it doesn't keep popping up telling you to login! (Unless you're working offline.)

 * 1. Responds to user input and grabs a new bearer token. -> Done
 * 1. Keeps the bearer token fresh and regfox status connected. -> Still to do

### DIFF
--- a/src/regdesk/login/login_manager.js
+++ b/src/regdesk/login/login_manager.js
@@ -1,0 +1,60 @@
+import { login } from '../../regfox/regfox_api.js';
+
+const STORAGE_LOCAL_KEY = 'regfox-bearer-details';
+
+/**
+ * Class has these main responsibilities.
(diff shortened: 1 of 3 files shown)


src/regdesk/login/login_manager.js
```diff
@@ -0,0 +1,60 @@
+import { login } from '../../regfox/regfox_api.js';
+
+const STORAGE_LOCAL_KEY = 'regfox-bearer-details';
```
Cellivar, line 3: Tempted to say these should live somewhere central. We can revisit it when it actually matters.

Cellivar (commented):
I like how clean this looks!

Cellivar (approved):

PR #39, opened 2022-02-03
Allow login form submission via enter key.

kobaj:
Does what it says on the tin! Not sure why exactly we need the keyup handler, but it won't work without it.


Cellivar (approved):
Fascinating, it being a form was insufficient? Maybe the enter key is being swallowed by something else? Probably fine as a workaround.

PR #40, opened 2022-02-04
Feature/kobaj login manager refresh

kobaj:
The code now grabs a new bearer token ~60 seconds before the previous one expires. Its not the prettiest logic, but I was trying to follow the pattern set about in the disable_auto_logout.js file. 

Since this is the first time calling the API from the frontend, turns out we needed to make a small adjustment to the export of the API too.

### DIFF
--- a/src/regdesk/login/login_manager.js
+++ b/src/regdesk/login/login_manager.js
@@ -8,6 +8,45 @@ const STORAGE_LOCAL_KEY = 'regfox-bearer-details';
  * 1. Keeps the bearer token fresh and regfox status connected.
  */
 export class LoginManager {
+  static #instance;
+  /** Only one instance of LoginManager can exist. */
+  constructor() {
+    if (LoginManager.#instance) {
+      throw new Error('LoginManager already has an instance!!!');
+    }
+    LoginManager.#instance = this;
+
+    setInterval(async () => {
+      try {
+        await this.refresh();
+      } catch (e) {
+        console.error('Failed to refresh bearer token, user will need to re-login.', e);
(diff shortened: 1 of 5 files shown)


src/regdesk/login/login_manager.js
```diff
@@ -8,6 +8,45 @@ const STORAGE_LOCAL_KEY = 'regfox-bearer-details';
  * 1. Keeps the bearer token fresh and regfox status connected.
  */
 export class LoginManager {
+  static #instance;
```
Cellivar, line 11: Interesting singleton hack, haha!

src/regdesk/login/login_manager.js
```diff
@@ -8,6 +8,45 @@ const STORAGE_LOCAL_KEY = 'regfox-bearer-details';
...
+      throw new Error('LoginManager already has an instance!!!');
+    }
+    LoginManager.#instance = this;
+
+    setInterval(async () => {
+      try {
+        await this.refresh();
+      } catch (e) {
+        console.error('Failed to refresh bearer token, user will need to re-login.', e);
+        await this.#setBearerDetails(null);
```
Cellivar, line 24: I wonder if we should add an event of some sort here that we could use to update the connected state? Or should this directly manage the "Connected" status in the top left?

Cellivar (approved):
Code LGTM! Just a question around how we'll manage the UI indicators for logged in status.

src/regdesk/login/login_manager.js
kobaj: Haha, thats my next pull request!

PR #41, opened 2022-02-04
Wire the Regfox Status to the login form.

kobaj:
I don't know how you make those cool gif screen recordings. So pretend this is neatly animated:

image


Cellivar (approved):
Nice work! Thanks for following the same event pattern, it's grown on me.

PR #42, opened 2022-02-06
Fix import paths for WebZLP

binaryf0x:
Importing from a library shouldn't reference the src directory.


kobaj (approved):
Yes! Good fix :D.

PR #43, opened 2022-02-07
Add login/logout buttons to regfox status dropdown.

kobaj:
Added buttons! Tried to follow the pattern we have with the printers where the buttons are consistent. Happy to switch it if you would prefer though.

image
image


Cellivar (dismissed):
A little concerned about the isLoggedIn() soup in regdesk.js, it's legible but takes a bit to think about. The rest looks good.

Cellivar (approved):
Better!

PR #44, opened 2022-02-08
Add a real search call to the communication manager.

kobaj:
It still needs a lot of work, but it does the search!

image

### DIFF
--- a/src/regdesk/communication_manager.js
+++ b/src/regdesk/communication_manager.js
@@ -4,6 +4,7 @@ import { TogglePaymentsBtn } from './toggle_payments_btn.js';
 
 import { MESSAGE_TYPE } from '../coms/communication.js';
 import { RegistrantDetails } from './registrant_details.js';
+import { RegfoxApi } from '../regfox/regfox_api.js';
 
 /**
  * Manager for handling communication to the extension and APIs.
@@ -12,6 +13,7 @@ export class CommunicationManager {
   #acceptingPayments;
   #togglePaymentsBtn;
   #printerMgr;
+  #loginMgr;
 
   /**
    * Indicates whether this terminal is accepting payments.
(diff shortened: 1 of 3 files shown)


src/regdesk/communication_manager.js
```diff
@@ -42,26 +46,26 @@ export class CommunicationManager {
...
+      .then((results) => Promise.all(results.map((result) => RegfoxApi.getRegistrationInfo(result.id, bearerToken))))
+      .then((results) => results.map((result) => new RegistrantDetails({
+        preferredName: result.preferredFirstName || result.name,
+        legalName: result.name,
+        birthdate: result.dateOfBirth,
+        amountDue: result.outstandingAmount100x, // FIXME there is a bug where #maybeSingleResultToUnpaid
+        // is unable to view this value, and therefor thinks everyone is unpaid.
+        badgeLine1: result.badgeLine1Text,
+        badgeLine2: result.badgeLine2Text || '',
+        badgeId: '12345678', // I have no idea which of id is this one.
```
Cellivar: The badgeID was the 8 digit value that was pulled out of the browser's URL, I'm not sure which actual API value this maps to.

src/regdesk/communication_manager.js
```diff
@@ -42,26 +46,26 @@ export class CommunicationManager {
...
-
-    return new Promise((resolve) => setTimeout(resolve, 500))
-      .then(() => arr) // The mock one!
-      .then(this.handleSearchResults.bind(this)); // The real one!
+  async startSearchForRegByName(searchString) {
+    // FIXME I think we still need to do some client side filtering for cancelled/duplicate/etc memberships.
+    const bearerToken = await this.#loginMgr.getBearerToken();
+    return RegfoxApi.searchRegistrations(searchString, bearerToken)
+      .then((result) => result.registrants.slice(0, 10)) // The next call hammers regfox, so we want to limit this as much as possible
+      .then((results) => Promise.all(results.map((result) => RegfoxApi.getRegistrationInfo(result.id, bearerToken))))
```
Cellivar, line 51: I guess there really isn't a lot of useful information to filter down first eh? Are we just generally hoping that the majority of searches looks for more specific results than 10 registrations?

src/regdesk/communication_manager.js
kobaj: Do you have an example URL, like from the test account? I can figure it out from there. :)

kobaj: Found one!

https://manage.webconnex.com/a/1311/reports/orders/23028022/details/26564608

The 26564608 is the BadgeId, and which is the same as `registrant.id`, will update!

src/regdesk/communication_manager.js
kobaj: Fair point, I think at the moment since we have no mechanism to inform the user "please search more specifically" then we can't include something like this.

Cellivar (approved):
Nice work!

PR #45, opened 2022-02-08
Bug/kobaj exchange token fails

kobaj:
I was adding an extra minute and occasionally having to re-login during my testing. Found the bug and fixed it!


Cellivar (approved):
Haha been there done that.

